PR #13, opened 2020-06-30
Make function definitions inline

neilweidinger:
Using the library in two separate source files was leading to multiple definition errors, as the `inline` specifier needs to be used in the definition rather than the declaration.

A few links that helped me understand how exactly `inline` quite works in C++ 😊 :
  - https://pabloariasal.github.io/2019/02/28/cpp-inlining/
  - https://isocpp.org/wiki/faq/inline-functions#where-to-put-inline-keyword

### DIFF
--- a/include/cpp-statsd-client/StatsdClient.hpp
+++ b/include/cpp-statsd-client/StatsdClient.hpp
@@ -36,25 +36,25 @@ public:
     //!@{
 
     //! Sets a configuration { host, port, prefix }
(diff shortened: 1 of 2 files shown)


vthiery:
Thanks for fixing this! I'll fix the CI first and then merge this

vthiery:
@neilweidinger I merged a fix for the CI. Could you please rebase your branch?

neilweidinger:
Awesome thank you, just rebased and should be good to go!

PR #16, opened 2021-01-11
fix: prevent batching thread to run indefinitely when UDPSender is destroyed

vthiery:
Make `m_mustExit` an `std::atomic<bool>` and use `std::memory_order_acq_rel` ordering so that other threads are guaranteed to see the change of value of `m_mustExit`.


jayv (approved):
Looking good, thx!

PR #17, opened 2021-05-27
Backport to c++11 and Add Back CI

kevinkreiser:
Hi! I'm not sure if you are accepting pull requests at the moment but I found your library and thought it looked useful. There were a couple of things stopping me from using it thought. The project I was looking to use it in is and open source routing engine which you can find here: https://github.com/valhalla/valhalla This routing engine is often run in a server environment and it would be great to collect metrics from it via statsd integration, which is what brought me here!

Sadly the project is targetting c++14 at the moment to maintain compatibility with older build systems with older compilers, usually for embedded systems like the Tesla in-car navigation system and other bespoke vehicle navigation systems. This meant that I needed to backport the code in a couple places (I chose to target c++11 since it has wide compatibility over the last decade). I'll make annotations to the diff to explain those changes.

The other thing I noticed is that this project is still targetting travis which semi-recently removed free support. So I've added github actions CI which builds and tests (though I haven't fully finished the testing) the code.

Let me know if some or any of these contributions would be welcome and thanks again for providing this library!

### DIFF
--- a/.github/workflows/linux.yaml
+++ b/.github/workflows/linux.yaml
@@ -24,5 +24,5 @@ jobs:
         run: |
           export LD_LIBRARY_PATH=.:$(cat /etc/ld.so.conf.d/* | grep -vF "#" | tr "\\n" ":" | sed -e "s/:$//g")
           #TODO: run statsd to collect these test metrics
-          testStatsdClient
+          ./testStatsdClient
           #TODO: validate test metrics received
--- a/.github/workflows/linux.yaml
+++ b/.github/workflows/linux.yaml
@@ -1,0 +1,28 @@
+name: linux
+on: [pull_request]
+
+jobs:
+  build:
+    name: build and test
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: dependencies
+        shell: bash
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y -qq make cmake gcc g++
+      - name: build
+        shell: bash
+        run: |
+          export LD_LIBRARY_PATH=.:$(cat /etc/ld.so.conf.d/* | grep -vF "#" | tr "\\n" ":" | sed -e "s/:$//g")
+          cmake . -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_SANITIZERS=On && make all -j$(nproc)
+      - name: test
+        shell: bash
+        run: |
+          export LD_LIBRARY_PATH=.:$(cat /etc/ld.so.conf.d/* | grep -vF "#" | tr "\\n" ":" | sed -e "s/:$//g")
+          #TODO: run statsd to collect these test metrics
+          testStatsdClient
+          #TODO: validate test metrics received
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 ![logo](https://raw.githubusercontent.com/vthiery/cpp-statsd-client/master/images/logo.svg?sanitize=true)
 
 [![Download](https://api.bintray.com/packages/vthiery/conan-packages/statsdclient%3Avthiery/images/download.svg)](https://bintray.com/vthiery/conan-packages/statsdclient%3Avthiery/_latestVersion)
-[![Build Status](https://travis-ci.org/vthiery/cpp-statsd-client.svg?branch=master)](https://travis-ci.org/vthiery/cpp-statsd-client)
+![Build Status](https://github.com/vthiery/cpp-statsd-client/actions/workflows/linux/badge.svg)
 [![Github Issues](https://img.shields.io/github/issues/vthiery/cpp-statsd-client.svg)](https://github.com/vthiery/cpp-statsd-client/issues)
 [![Average time to resolve an issue](http://isitmaintained.com/badge/resolution/vthiery/cpp-statsd-client.svg)](http://isitmaintained.com/project/vthiery/cpp-statsd-client "Average time to resolve an issue")
 
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 ![logo](https://raw.githubusercontent.com/vthiery/cpp-statsd-client/master/images/logo.svg?sanitize=true)
 
 [![Download](https://api.bintray.com/packages/vthiery/conan-packages/statsdclient%3Avthiery/images/download.svg)](https://bintray.com/vthiery/conan-packages/statsdclient%3Avthiery/_latestVersion)
-![Build Status](https://github.com/vthiery/cpp-statsd-client/actions/workflows/linux/badge.svg)
+[![Build Status](https://github.com/vthiery/cpp-statsd-client/actions/workflows/linux/badge.svg)](https://github.com/vthiery/cpp-statsd-client/actions?workflow=linux)
 [![Github Issues](https://img.shields.io/github/issues/vthiery/cpp-statsd-client.svg)](https://github.com/vthiery/cpp-statsd-client/issues)
 [![Average time to resolve an issue](http://isitmaintained.com/badge/resolution/vthiery/cpp-statsd-client.svg)](http://isitmaintained.com/project/vthiery/cpp-statsd-client "Average time to resolve an issue")
 
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ The batchsize is the only parameter that cannot be changed for the time being.
 The client supports batching of the metrics:
 
 - the unit of the batching parameter is bytes,
-- it is optional and not setting it will result in an immediate send of any metrics,
+- it is optional and passing a 0 will result in an immediate send of any metrics,
 
 If set, the UDP sender will spawn a thread sending the metrics to the server every 1 second by aggregates. The aggregation logic is as follows:
 
--- a/include/cpp-statsd-client/StatsdClient.hpp
+++ b/include/cpp-statsd-client/StatsdClient.hpp
@@ -3,6 +3,7 @@
 
 #include <cstdlib>
 #include <string>
+#include <memory>
 #include "UDPSender.hpp"
 
 namespace Statsd {
@@ -29,15 +30,18 @@ public:
                  const std::string& prefix,
                  const uint64_t batchsize = 0) noexcept;
 
+    StatsdClient(const StatsdClient&) = delete;
+    StatsdClient& operator=(const StatsdClient&) = delete;
+
     //!@}
 
     //!@name Methods
     //!@{
 
-    //! Sets a configuration { host, port, prefix }
-    void setConfig(const std::string& host, const uint16_t port, const std::string& prefix) noexcept;
+    //! Sets a configuration { host, port, prefix, batchsize }
+    void setConfig(const std::string& host, const uint16_t port, const std::string& prefix, const uint64_t batchsize = 0) noexcept;
 
-    //! Returns the error message as an optional std::string
+    //! Returns the error message as an std::string
     const std::string& errorMessage() const noexcept;
 
     //! Increments the key, at a given frequency rate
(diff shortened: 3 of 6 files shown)


.github/workflows/linux.yaml
```diff
@@ -0,0 +1,28 @@
...
+        run: |
+          export LD_LIBRARY_PATH=.:$(cat /etc/ld.so.conf.d/* | grep -vF "#" | tr "\\n" ":" | sed -e "s/:$//g")
+          cmake . -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_SANITIZERS=On && make all -j$(nproc)
+      - name: test
+        shell: bash
+        run: |
+          export LD_LIBRARY_PATH=.:$(cat /etc/ld.so.conf.d/* | grep -vF "#" | tr "\\n" ":" | sed -e "s/:$//g")
+          #TODO: run statsd to collect these test metrics
+          ./testStatsdClient
+          #TODO: validate test metrics received
```
kevinkreiser, line 28: i think we should flesh this bit out a little and actually run statsd and check that it got all the test metrics. im new to statsd but if we could figure something to just dump them to file it would be pretty straight forward

README.md
```diff
@@ -3,7 +3,7 @@
 ![logo](https://raw.githubusercontent.com/vthiery/cpp-statsd-client/master/images/logo.svg?sanitize=true)
 
 [![Download](https://api.bintray.com/packages/vthiery/conan-packages/statsdclient%3Avthiery/images/download.svg)](https://bintray.com/vthiery/conan-packages/statsdclient%3Avthiery/_latestVersion)
-[![Build Status](https://travis-ci.org/vthiery/cpp-statsd-client.svg?branch=master)](https://travis-ci.org/vthiery/cpp-statsd-client)
+![Build Status](https://github.com/vthiery/cpp-statsd-client/actions/workflows/linux/badge.svg)
```
kevinkreiser, line 6: this will be the badge from the github actions run. as far as i could tell they dont have a link to the latest build though to see like a log and details. seems each "run" gets a hash assigned to it so there is no human readable way to tell what the link should be here

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -28,7 +27,7 @@ class StatsdClient {
     StatsdClient(const std::string& host,
                  const uint16_t port,
                  const std::string& prefix,
-                 const std::optional<uint64_t> batchsize = std::nullopt) noexcept;
```
kevinkreiser, line 31: there were two main changes for c++11 compatibility. the first was removing the use of `std::optional`. I chose to use a sentinel value instead. For the batchsize i just used `max` of `uint64_t` to signal unset.

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -39,7 +38,7 @@ class StatsdClient {
     void setConfig(const std::string& host, const uint16_t port, const std::string& prefix) noexcept;
 
     //! Returns the error message as an optional std::string
-    std::optional<std::string> errorMessage() const noexcept;
+    const std::string& errorMessage() const noexcept;
```
kevinkreiser, line 41: for strings, i used the empty string to signal unset.

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -185,28 +184,26 @@ inline void UDPSender::send(const std::string& message) noexcept {
         if (m_batchingMessageQueue.empty() || m_batchingMessageQueue.back().length() > m_batchsize) {
             m_batchingMessageQueue.push_back(message);
         } else {
-            std::rbegin(m_batchingMessageQueue)->append("\n").append(message);
```
kevinkreiser, line 188: this was another back port. `std::rbegin` as added in a newer revision

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -185,28 +184,26 @@ inline void UDPSender::send(const std::string& message) noexcept {
...
-
     if (m_isInitialized) {
         return true;
     }
 
     // Connect the socket
     m_socket = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
     if (m_socket == -1) {
-        m_errorMessage = "Could not create socket, err="s + std::strerror(errno);
+        m_errorMessage = std::string("Could not create socket, err=") + std::strerror(errno);
```
kevinkreiser, line 206: this was the only place that was making use of `string_literals` in the other cases there was already an `std::string` as one of the operands so `const char*` worked fine

tests/testStatsdClient.cpp
```diff
@@ -11,27 +11,27 @@ int main() {
 
     // Increment "coco"
     client.increment("coco");
-    assert(!client.errorMessage());
+    assert(client.errorMessage().empty());
```
kevinkreiser, line 14: here we check the sentinel value `empty` instead of checking if its "unset"

vthiery:
Hi @kevinkreiser 

PRs are vm welcome so thanks for the work you've already put in!

Backporting to C++11 seems like a good idea. It'd make the client available to more users at little cost (as you showed, there aren't much C++14/17 features being used).
Also, the move to Github Actions is a long-due item that I never had the time to do, so feel free to change this as well!

README.md
vthiery: Something like `https://github.com/vthiery/cpp-statsd-client/actions?workflow=linux` should work

.github/workflows/linux.yaml
vthiery: The current test suite is very shallow indeed. I won't have the time to setup better tests now but this is definitely something worth working on. 
If you already have good ideas, feel free to suggest in another PR or raise an issue.

include/cpp-statsd-client/StatsdClient.hpp
vthiery: Do you see anything against using `0` as the unset signal?

.github/workflows/linux.yaml
kevinkreiser: I'd be happy to contribute them as well in the next PR!

include/cpp-statsd-client/StatsdClient.hpp
kevinkreiser: nope, no reason not to use that, sounds good!

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -116,21 +115,21 @@ class UDPSender final {
     //!@}
 
     //! Error message (optional string)
-    std::optional<std::string> m_errorMessage;
+    std::string m_errorMessage;
```
kevinkreiser, line 118: on each call should we be resetting this to the empty string? it seems like previous behavior remembered whatever the last error was no matter how long ago it happened. is that desired?

kevinkreiser:
@vthiery im very happy to contribute! i've gone over your review comments and made the updates. please let me know if there is anything more to do!

.github/workflows/linux.yaml
kevinkreiser: i see that statsd has a "console" backend which should write to file if we run it correctly. i think this would allow us to do a proper "integration" test in CI. we should add unit tests too of course

vthiery: We can discuss this further in the next PR but running something like this with docker-compose and calling the graphite API should be simple enough:

```
gostatsd:
  image: atlassianlabs/gostatsd:29.0.0
  command: gostatsd --config-path=/etc/gostatsd/config.toml
  links:
    - graphite
  ports:
    - "8125:8125/udp"
    - "8126:8126"
    - "8181:8181"
  volumes:
    - ${PWD}:/etc/gostatsd:ro

graphite:
  image: sitespeedio/graphite
  ports:
    - "8080:80"
    - "8000:8000"
    - "2003:2003"
    - "2004:2004"
    - "7002:7002"
```

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -39,7 +38,7 @@ class StatsdClient {
     void setConfig(const std::string& host, const uint16_t port, const std::string& prefix) noexcept;
 
     //! Returns the error message as an optional std::string
```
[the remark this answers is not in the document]
kevinkreiser: haha i did this one last night while i was tidying up some more. will push it up shortly

include/cpp-statsd-client/UDPSender.hpp
kevinkreiser: did this one last night too

.github/workflows/linux.yaml
kevinkreiser: interesting! i was able to loaded up statsd directly on ubuntu and configure the console backend and see that it spit out the various stats that the test program sent, in javascript/json notation. i didnt update the CI to check it though. would you prefer graphite? and if so, how does one retrieve the values then to make sure they are what they should be?

include/cpp-statsd-client/StatsdClient.hpp
kevinkreiser: done!

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -67,25 +70,25 @@ class StatsdClient {
...
+    : m_prefix(prefix), m_sender(new UDPSender{host, port, batchsize}) {
+    // Initialize the random generator to be used for sampling
     std::srand(time(nullptr));
 }
 
-inline void StatsdClient::setConfig(const std::string& host, const uint16_t port, const std::string& prefix) noexcept {
+inline void StatsdClient::setConfig(const std::string& host, const uint16_t port, const std::string& prefix, const uint64_t batchsize) noexcept {
     m_prefix = prefix;
-    m_sender.setConfig(host, port);
+    m_sender.reset(new UDPSender(host, port, batchsize));
```
kevinkreiser, line 87: so i changed this bit of the underlying implementation an i'm not sure what you'll think of it so please let me know.

initialization of the socket was done lazily before. basicaly when you wanted to send something it would first check if you've initialized, until that point it wouldnt bother with initializing. this causes a couple of problems:

1. when testing you basically cant tell that anything happened when you construct the object, initializing right away allows you to fail fast and detect it (and error or handle the error)
2. there is a race between `sendToDaemon` and `setConfig` the sender. the latter and tear down the socket while the former is in the middle of sending one of the messages batching mode (ie from the batching thread). we could have fixed this with synchronization but it seemed easier to just initialize once and let that be the config for the life time of the object. the top level interface still allows you to "re-configure" but under the hood it just makes a new `UDPSender`

tests/testStatsdClient.cpp
```diff
@@ -1,37 +1,65 @@
...
-    client.setConfig("localhost", 8000, "myPrefix.");
+    client.setConfig("localhost", 8125, "myPrefix.");
+    if (!client.errorMessage().empty()) {
+        std::cerr << client.errorMessage() << std::endl;
+        return EXIT_FAILURE;
+    }
 
     // Increment "coco"
     client.increment("coco");
-    assert(!client.errorMessage());
```
kevinkreiser, line 14: i figured that also release builds should fail the tests so rather than going with `asserts` i just check the values and return failure.

README.md
kevinkreiser: ive done this!

include/cpp-statsd-client/StatsdClient.hpp
vthiery: Hmm I can't remember why I went for lazy initialization back then :thinking: 
I'm fine changing this since it makes the code simpler and easier to test.

Using the `unique_ptr` and `reset` is the right thing to do, thanks :+1:

.github/workflows/linux.yaml
vthiery: I would have used [Graphite's HTTP API](https://graphite-api.readthedocs.io/en/latest/api.html#metrics-find) to check that the metrics were received but I'm fine with using only statsd.
Having this running in the CI is probably worth another PR though.

kevinkreiser:
also i noticed there is a style here with the comments on the methods. its is unfamiliar to me can you mention what its called so i can look it up? are there any other things i need to tidy up before this is ready to be merged?

vthiery:
> also i noticed there is a style here with the comments on the methods. its is unfamiliar to me can you mention what its called so i can look it up?

The documentation is following the [Doxygen](https://www.doxygen.nl/index.html) style (even though I don't generate the documentation).

> are there any other things i need to tidy up before this is ready to be merged?

I think it's ready to be merged.
Since the tests are rather limited, I would prefer to wait for the addition of more tests before cutting another release. Would that be ok or do you need an official release to adopt it in [valhalla](https://github.com/valhalla/valhalla)?

kevinkreiser:
no release is needed for us, we will use it as a submodule. before we actually use it we probably want to make a couple other changes too (i know support for tags was one of the features we'd like to add). i will work on the tests and support for tags in the very near (next few days) future.

vthiery (approved):

vthiery:
Thanks a lot for the work @kevinkreiser !

PR #19, opened 2021-06-07
End to End Testing

### LINKED ISSUE #18: Add Integration Tests
Recently we added github actions to the repo which build and run a test application that publishes stats a few different ways. At the moment all it does is test that the code compiles and doesn't complain when you run it. This is simply not thorough enough. For proper testing we need to:

1. enable linux github action and set it as required (@vthiery can you help me out here?)
2. run statsd with either the graphite or console backend to collect stats from our test program
3. after the test program finishes verify that the stats actually landed at the backend

There may be some interesting (ie more difficult) stuff to test here regarding randomness and the frequency of collection, but I'm sure we can figure something out!

@vthiery i suppose you cannot assign this task to me but suffice it to say i will PR this work

kevinkreiser:
fixes #18 

This PR expands the test application that we run in CI so that it spawns a thread which listens on the statsd UDP port and stores the messages it got in memory. Then we validate that the messages it received were the ones we expected it to.

I made quite a couple changes here to make this more testable:

1. Change the code to use c++11 style random number generation and add seeding of the RNG to the client interface (this allows repeatable results in the test)
2. Add a mock statsd server implementation to the test directory which we throw in a background thread to collect the messages it receives for validation

There were other changes I made as nice-to-haves:

1. Rather than using a char[] buffer I use a string directly, this will allow us to avoid copying the buffer into a string when it goes on to the UDPSender, in the future we can make this more efficient by making it a member and keeping it a fixed size
2. I put a bunch of `TODO`s as questions in the code or suggestions, please let me know which ones i should implement or remove

Another side question I have is, should we add some semver to the API? I've been changing the main methods here and there and whenever you make your next release I suppose the version should reflect that. I think having the version number in code might be helpful to others as well.


CMakeLists.txt
```diff
@@ -10,7 +10,10 @@ include_directories(${CPP_STATSD_CLIENT_DIR})
 install(DIRECTORY ${CPP_STATSD_CLIENT_DIR} DESTINATION include)
 
 find_package(Threads)
-add_executable(testStatsdClient ${CMAKE_CURRENT_SOURCE_DIR}/tests/testStatsdClient.cpp)
+add_executable(testStatsdClient 
+  ${CMAKE_CURRENT_SOURCE_DIR}/tests/testStatsdClient.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/tests/StatsdServer.hpp)
+target_include_directories(testStatsdClient PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/tests)
```
kevinkreiser, line 16: making sure the test can actually see and use the mock statsd server implementation.

tests/StatsdServer.hpp
```diff
@@ -0,0 +1,60 @@
+#ifndef STATSD_SERVER_HPP
+#define STATSD_SERVER_HPP
+
+// TODO: should we actually add this to the distributed headers? With a tiny bit of work
+//  someone could write their own c++ statsd server
+
+
+class StatsdServer {
```
kevinkreiser, line 8: so originally i started testing by making statsd a submodule to the repo and was going to actually run it with node etc but there were a couple things i didnt like about it:

- i have to install node just to test it, this takes a bunch of time in CI and can go down from time to time
- the testing methodology is messy. 
  - you have to start up statsd and possibly another backend in the background and then kill them at some point
  - you have to collect the results either in a log file or in a database and then pull that out and transcode it to something you can easily validate
  - validation has to be done with yet another script

so instead i opted to do a much cleaner test all in process. the statsd format is so simple it didnt make sense to do all of this other complicated stuff when i can just write a tiny bit of code to suck bytes out of a UDP socket and compare them so thats what i did. this also has the benefit of allowing downstream users to test that their own application works as they expect in terms of reporting metrics.

tests/StatsdServer.hpp
```diff
@@ -0,0 +1,60 @@
+#ifndef STATSD_SERVER_HPP
+#define STATSD_SERVER_HPP
+
+// TODO: should we actually add this to the distributed headers? With a tiny bit of work
+//  someone could write their own c++ statsd server
```
kevinkreiser, line 5: should i move this to the include directory so that others can use it, i will personally use this in my project since it lets me validate that im sending the stats that i expect to.

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -63,6 +64,9 @@ class StatsdClient {
     void send(const std::string& key, const int value, const std::string& type, const float frequency = 1.0f) const
         noexcept;
 
+    //! Seed the RNG that controls sampling
+    void seed(unsigned int seed = std::random_device()()) noexcept;
```
kevinkreiser, line 68: this is critical to allow for testing of the sampling rate function calls

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -71,15 +75,27 @@ class StatsdClient {
 
     //! The UDP sender to be used for actual sending
     std::unique_ptr<UDPSender> m_sender;
+
+    //! The random number generator for handling sampling
+    static thread_local std::mt19937 m_random_engine;
```
kevinkreiser, line 80: its expensive to make these over and over, we could have one per object and that would be fine but since its thread local static can work in this case. let me know if youd like me to just make it a plain instance member variable though

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -71,15 +75,27 @@ class StatsdClient {
...
                                   const uint16_t port,
                                   const std::string& prefix,
                                   const uint64_t batchsize) noexcept
     : m_prefix(prefix), m_sender(new UDPSender{host, port, batchsize}) {
+    // TODO: the final metric strings do not automatically place a '.' between the prefix and the key
+    //  This might be unexpected to end users. We should either document this or we could just add the
+    //  automatic '.' when the prefix is non empty
+    //if(!m_prefix.empty()) {
+    //    m_prefix.push_back('.');
+    //}
```
kevinkreiser, line 95: it seemed a bit odd that we dont do this automatically. do you feel like we should or that we should document that its not automatic?

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -116,33 +132,52 @@ inline void StatsdClient::send(const std::string& key,
...
-        return std::fabs(frequency - 1.0f) < epsilon;
-    };
-
-    // Test if one should send or not, according to the frequency rate
-    if (!isFrequencyOne(frequency)) {
-        if (frequency < static_cast<float>(std::rand()) / RAND_MAX) {
-            return;
-        }
+    constexpr float epsilon{0.0001f};
+    const bool isFrequencyOne = std::fabs(frequency - 1.0f) < epsilon;
```
kevinkreiser, line 136: frequency is constant, which means theres no point in evaluating if its `1.f` more than once. so it was just cleaner to remove the lambda and store the boolean to use over and over

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -116,33 +132,52 @@ inline void StatsdClient::send(const std::string& key,
...
+
+    // A valid metric must be at least 6 chars in length
+    if(string_len < 6){
+        //TODO: we dont have access to the error message here
+        return;
     }
 
+    // Trim the trailing null chars
+    // TODO: perf improvement: send the len along and keep the buffer around as a member variable
+    buffer.resize(std::min(string_len, static_cast<int>(buffer.size())));
```
kevinkreiser, line 171: so i noticed that we were copying a `char[]` into an `std::string` when we call `send` below. i did a small refactor to just not copy but rather start with a string in the first place. long term i think we should make the buffer a member variable and keep it a fixed size and then send it and its `snprintf`'d length to the send funciton so that it can avoid copying all together (except in the case it has to be batched).

i want to make another pr where refactor the batching a bit to maybe keep a single buffer with all the unsent stuff in it (either via the background thread or not) and just flush that from time to time. this would avoid any unneeded reallocs or deallocs (except when we get over the UDP limit which is like 64k)

tests/testStatsdClient.cpp
```diff
@@ -1,18 +1,49 @@
 #include <StatsdClient.hpp>
-#include <cassert>
+#include <StatsdServer.hpp>
 #include <iostream>
 
 using namespace Statsd;
 
+// We just keep storing metrics in an vector until we hear a special one, then we bail
```
kevinkreiser, line 7: so basically we spawn a thread in the background which makes a server to listen to our client. everytime the server gets a message it parses it (splitting it on endline characters) and pushes it into the vector it was provided. once it hears a special message with the keyword `DONE` in it it quits

tests/testStatsdClient.cpp
```diff
@@ -34,6 +65,7 @@ int main(int argc, char** argv) {
     }
 
     // Adjusts "toto" by +3
+    client.seed(19); // this seed gets a hit on the first call
```
kevinkreiser, line 68: for messages sent with sampling set to less than `1.f` we modify the RNG's seed to make sure it will be sent

tests/testStatsdClient.cpp
```diff
@@ -48,18 +80,49 @@ int main(int argc, char** argv) {
...
 
+    // TODO: should sampling rates above 1 error or be capped?
     // Send a metric explicitly
     client.send("tutu", 4, "c", 2.0f);
     if (!client.errorMessage().empty()) {
         std::cerr << client.errorMessage() << std::endl;
         return EXIT_FAILURE;
     }
 
+    // TODO: reconfigure and test batching deamon
```
kevinkreiser, line 98: ill finish this TODO in this PR

tests/testStatsdClient.cpp
```diff
@@ -1,65 +1,150 @@
...
+    // If any of these tests fail they throw an exception, not catching makes for a nonzero return code
+
+    // general things that should be errors
+    testErrorConditions();
+    // reconfiguring how you are sending
+    testReconfigure();
+    // no batching
+    testSendRecv(0);
+    // background batching
+    testSendRecv(4);
```
kevinkreiser, line 146: these are the different sets of tests i came up with. i should probably enable code coverage to see what is and isnt covered at the moment, but i'm hoping its getting close to 100% (other than the stuff in the unit test code that arent covered because we arent actualy hitting failures in the tests haha)

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -116,33 +132,52 @@ inline void StatsdClient::send(const std::string& key,
...
+                                          m_prefix.c_str(),
+                                          key.c_str(),
+                                          value,
+                                          type.c_str(),
+                                          frequency);
     }
 
+    // A valid metric must be at least 4 chars in length ie :n|c
+    if (string_len < 4) {
+        //TODO: we dont have access to the error message here
```
kevinkreiser, line 165: i feel like this and other scenarios make exceptions feel pretty natural for error handling in this library. we could make a small struct that holds a numeric error number and the string and throw that on error rather than have the caller always check the result. i'm just wondering if maybe you had a specific use-case in mind that maybe `noexcept` were important?

though for this particular error i can remove it completely if i can refactor how the type is selected.. let me work on that

vthiery (commented):
This is awesome!

If possible, I would keep the scope to testing only and avoid breaking the API here. We can (should?) break it in another set of PR though.

kevinkreiser:
@vthiery thanks for all the review feedback i'll get them all changed and break out the other stuff into another PR. should have that doen shortly!

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -39,29 +44,34 @@ class StatsdClient {
     //!@{
 
     //! Sets a configuration { host, port, prefix, batchsize }
-    void setConfig(const std::string& host, const uint16_t port, const std::string& prefix, const uint64_t batchsize = 0) noexcept;
+    void setConfig(const std::string& host,
+                   const uint16_t port,
+                   const std::string& prefix,
+                   const uint64_t batchsize = 0) noexcept;
```
kevinkreiser, line 50: i dont believe i've changed this line, yet it seems the linter decided to. in other projects ive learned that i had to pin the version of clang-format because different versions of that applied the style differently and so people would be committing constant lint changes back and fourth. i can make a separate PR to pin clang-format and run lint as a part of the github actions

kevinkreiser:
i believe ive removed some of the additional changes here and have just the changes discussed in the first round of review:

1. random number gen change to allow for testing
2. cap frequency to valid range
3. vastly expanded testing
4. help methods for determining an unusable socket (invalid file descriptor)

kevinkreiser:
@vthiery i got master merged in with the linting PR, i think this is ready for review again

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -117,41 +125,61 @@ inline void StatsdClient::gauge(const std::string& key,
...
-        }
+                               float frequency) const noexcept {
+    // Bail if we can't send anything anyway
+    if (!m_sender->initialized()) {
+        return;
+    }
+
+    // A valid frequency is: 0 <= f <= 1
+    // At 0 you never emit the stat, at 1 you always emit the stat and with anything else you roll the dice
+    frequency = std::max(std::min(frequency, 1.f), 0.f);
```
kevinkreiser, line 143: i made `frequency` non-const so i could do this modification here, if you'd prefer it to stay `const` just let me know!

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -230,6 +238,10 @@ inline void UDPSender::sendToDaemon(const std::string& message) noexcept {
     }
 }
 
+inline bool UDPSender::initialized() const noexcept {
```
kevinkreiser, line 241: its too bad the UDP server and client cant share these bits of code, but at the very least the work the same way even if the code is duplicated

README.md
```diff
@@ -65,6 +64,38 @@ int main()
 
     // Send a metric explicitly
     client.send("tutu", 4, "c", 2.0f);
+    exit(0);
+}
+```
+
+### Advanced Testing
+
+A simple mock StatsD server can be found at `tests/StatsdServer.hpp`. This can be used to do simple validation of your application's metrics, typically in the form of unit tests. In fact this is the primary means by which this library is tested. The mock server itself is not distributed with the library so to use it you'd need to vendor this project into your project. Once you have though, you can test your application's use of the client like so:
```
kevinkreiser, line 73: here ive implemented the first option you suggested which was to just document the existance of the mock server.

a second option to avoid having to distribute the mock server as another header would be to put it into the `UDPSender.hpp` header and have it share some of its common bits with it the sender. this would both reduce duplicated code and make the mock receiver accesable to downstream projects.

the third option would be to hide it completely and remove this entry from the readme and just have it defined and used in the test applications code.

@vthiery let me know what you like best of the 3 options

include/cpp-statsd-client/StatsdClient.hpp
kevinkreiser: makes sense to me!

tests/testStatsdClient.cpp
```diff
@@ -1,65 +1,154 @@
...
-    // Record a timing of 2ms for "myTiming"
-    client.timing("myTiming", 2, 0.1f);
-    if (!client.errorMessage().empty()) {
-        std::cerr << client.errorMessage() << std::endl;
-        return EXIT_FAILURE;
+    StatsdClient client("localhost", 8125, "first.");
+    client.send("foo", 1, MetricType::COUNTER, 1.f);
+    if(server.receive() != "first.foo:1|c")
+        throw std::runtime_error("Incorrect stat received");
+
```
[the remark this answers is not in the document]
kevinkreiser: i dont know if there is a style rule to force the braces, but it seems the current style doesnt do that. for me it seems to, if it could fit it, put it on one line. i figure you'd rather update the style in another pr to handle adding braces anyway though

tests/testStatsdClient.cpp
```diff
@@ -1,65 +1,154 @@
...
+        throwOnError(client);
+        expected.emplace_back("sendRecv.titi:3|g");
+
+        // Record a timing of 2ms for "myTiming"
+        client.seed(19);
+        client.timing("myTiming", 2, 0.1f);
+        throwOnError(client);
+        expected.emplace_back("sendRecv.myTiming:2|ms|@0.10");
+
+        // TODO: should sampling rates above 1 error or be capped?
```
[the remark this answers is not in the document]
kevinkreiser: done

kevinkreiser:
OK I feel pretty sure now that I've covered all of the things discussed in the review comments. Sorry for taking such a long time on that one!

README.md
vthiery: I prefer option one for the time being. Thanks for the doc!

include/cpp-statsd-client/StatsdClient.hpp
vthiery: No I guess it's better to remove the `const` keyword if we modify it internally, it makes the intent clear.

vthiery (approved):
LGTM

Again, thanks a lot for all the great work <3 !

PR #24, opened 2021-06-09
ci: add lint workflow

### LINKED ISSUE #22: Cleanup: Run Linter in CI
Currently CI does not check for properly linted code. We use clang-format here with the checked in style but we dont validate it on CI which makes for a bunch of work for any reviewer to sift through. To save some work, lets have the computers do it for us and run lint and fail the build if doing so shows any differences with what is checked in.

Also differences in versions of clang-format will do different things with your style file. Should we pin the version to avoid that problem?

vthiery:
Fixes #22 

* add lint workflow + specify clang-format version
* lint the code
* remove deprecated travis CI file
* remove useless git submodule file
* rename test workflow and trigger on push + pull request
* fix badges in readme


kevinkreiser (approved):
Looks great! I wonder what the output of that action looks like when the lint fails.. https://github.com/vthiery/cpp-statsd-client/runs/2781117211  cool it shows the actual diff! that is perfect!

PR #25, opened 2021-06-09
Update Cmake to Make Library Easier to Use in Downstream Projects

### LINKED ISSUE #23: Cleanup: CMake Doesn't Export the Library as an Interface
At the moment the cmake setup we have works to build the test program but there are a couple of minor issues with it:

1. the test program doesnt `depend` on the library (because there is no library target) and so you can get in a situation where you make changes to the header and the test program isnt able to tell it needs to be rebuilt
2. depending on how someone wants to use this library in a downstream project they may need to do some tricks. we should create a library target and mark it as an interface so that downstream users via cmake will know where to look for includes (this includes the test executable as well)
3. we dont set a version or a description
4. other little stuff?

kevinkreiser:
fixes #23 

The primary changes here are:

1. make an actual library target as an interface so that downstream targets can depend on it
2. fill out the basic project info
3. install package config files for cmake discoverability
4. fix include directives so that installations of the library will work properly (ie in the case someone installs rather than vendors the library)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@ set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_EXTENSIONS OFF)
 include(GNUInstallDirs)
 include(CMakePackageConfigHelpers)
-
+find_package(Threads)
 
 # The library target
 add_library(${PROJECT_NAME} INTERFACE)
@@ -17,6 +17,7 @@ target_include_directories(
   ${PROJECT_NAME}
   INTERFACE $<BUILD_INTERFACE:${${PROJECT_NAME}_SOURCE_DIR}/include>
             $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
+target_link_libraries(${PROJECT_NAME} INTERFACE Threads::Threads)
 
 # The installation and pkg-config-like cmake config
 install(TARGETS ${PROJECT_NAME}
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -1,19 +1,51 @@
...
+  INSTALL_DESTINATION
+  ${CMAKE_INSTALL_DATAROOTDIR}/${PROJECT_NAME}/cmake)
+install(EXPORT ${PROJECT_NAME}_Targets
+        FILE ${PROJECT_NAME}Targets.cmake
+        NAMESPACE ${PROJECT_NAME}::
+        DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/${PROJECT_NAME}/cmake)
+install(FILES "${PROJECT_BINARY_DIR}/${PROJECT_NAME}Config.cmake"
+              "${PROJECT_BINARY_DIR}/${PROJECT_NAME}ConfigVersion.cmake"
+        DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/${PROJECT_NAME}/cmake)
+install(DIRECTORY ${PROJECT_SOURCE_DIR}/include DESTINATION include)
```
kevinkreiser, line 42: when our headers land on the file system the default location will be `/usr/include/cpp-statsd-client/*.hpp`. I think the namespacing (ie subfolder) is a good thing as we have more than one header with names that dont share a common prefix. so organizationally i think this is the best option

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -1,10 +1,10 @@
 #ifndef STATSD_CLIENT_HPP
 #define STATSD_CLIENT_HPP
 
+#include <cpp-statsd-client/UDPSender.hpp>
```
kevinkreiser, line 4: matching the installation which has the subdirectory and using the `<>` brackets to search the system include paths first

tests/testStatsdClient.cpp
```diff
@@ -1,5 +1,5 @@
-#include <StatsdClient.hpp>
 #include <cassert>
+#include <cpp-statsd-client/StatsdClient.hpp>
```
kevinkreiser, line 2: this one can probably stay `""` i'll revert that

CMakeLists.txt
```diff
@@ -1,19 +1,51 @@
...
 
+# The test targets
 find_package(Threads)
 add_executable(testStatsdClient ${CMAKE_CURRENT_SOURCE_DIR}/tests/testStatsdClient.cpp)
-target_link_libraries(testStatsdClient ${CMAKE_THREAD_LIBS_INIT})
+target_link_libraries(testStatsdClient ${CMAKE_THREAD_LIBS_INIT} ${PROJECT_NAME})
```
kevinkreiser, line 47: having the test target depend on the library target makes it so that modifying the latter will trigger recompilation of the the former. previously you could get yourself in a state where changes wouldnt make the test target need to be rebuilt

kevinkreiser:
@vthiery this is ready for review!

kevinkreiser:
@vthiery is there anything else i should do here to get this ready for review? let me know if anything is missing!

vthiery (approved):
Thanks for the work @kevinkreiser !

I tried to follow but as I said, I'm quite rusty with `cmake` so I'll have to trust you on this one.

vthiery:
> @vthiery is there anything else i should do here to get this ready for review? let me know if anything is missing!

I was reviewing it already, sorry for the delay

kevinkreiser:
@vthiery no problem at all, thank you so much for the review :bow:  i should have another PR for you shortly tonight or tomorrow morning!

Issue #27, opened 2021-06-09
Feat: Automatically Add "." Between Prefix and Metric Key


kevinkreiser:
At the moment the API requires you to specify your metric prefix with the trailing `.` OR to specify your metrix key with a leading `.` to separate the two in the outgoing metric. We should do this convenience for people and document that it is being done. More info here: https://github.com/vthiery/cpp-statsd-client/pull/19#discussion_r647171349

PR #28, opened 2021-06-15
handle the dot between prefix and key automatically

kevinkreiser:
fixe #27 

this also updates doc strings and README exmaples and since it is API breaking it adds version macros. let me know if i should remove the macros, i know we didnt really discuss them before.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # Basic project setup
 cmake_minimum_required(VERSION 3.5)
 project(cpp-statsd-client
-        VERSION 1.0.2
+        VERSION 2.0.0
         LANGUAGES CXX
         DESCRIPTION "A header-only StatsD client implemented in C++"
         HOMEPAGE_URL "https://github.com/vthiery/cpp-statsd-client")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # Basic project setup
 cmake_minimum_required(VERSION 3.5)
 project(cpp-statsd-client
-        VERSION 2.0.0
+        VERSION 1.0.2
         LANGUAGES CXX
         DESCRIPTION "A header-only StatsD client implemented in C++"
         HOMEPAGE_URL "https://github.com/vthiery/cpp-statsd-client")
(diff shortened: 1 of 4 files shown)


vthiery:
Thanks for the patch!

Regarding the macros, what's their use exactly?
I intend to cut the next release only when all the changes I attached to the `v2` milestone will be done.
I'll only update the version in the cmake file, tag and release, so unless there's a good reason, we should avoid setting these macros.

kevinkreiser:
> what's their use exactly?

they allow downstream library maintainers to write protective code around API changes so that their library can work on whatever target platform they are being compiled. you might see something like this in downstream code:

```c++
// newer versions handle the dot automatically
#if STATSD_CLIENT_VERSION_MAJOR >= 2
StatsdClient client{ "127.0.0.1", 8080, "prefix", 20 };
// older versions require you to put the dot manually
#else
StatsdClient client{ "127.0.0.1", 8080, "prefix.", 20 };
#endif
```

this is probably less useful for a header only library where people are usually vendoring it rather than getting it as an installed package on their platform. but if the preferred way of getting this library is to install a package then the downstream library that depends on it might do like above to make sure whatever version they got from the package manager still worked in their code.

no problem if you think its not needed, i dont feel strongly about it, i can happily remove it!

tests/testStatsdClient.cpp
```diff
@@ -112,9 +123,9 @@ void testSendRecv(uint64_t batchSize) {
         expected.emplace_back("sendRecv.myTiming:2|ms|@0.10");
 
         // Send a metric explicitly
-        client.send("tutu", 4, "c", 2.0f);
+        client.send("tutu", 241, "c", 2.0f);
```
kevinkreiser, line 126: this was just because i thought it would be useful to exorcise some bigger numbers some times

CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # Basic project setup
 cmake_minimum_required(VERSION 3.5)
 project(cpp-statsd-client
-        VERSION 1.0.2
+        VERSION 2.0.0
```
vthiery, line 4: Can we wait for a release to be prepared?

When we're ready to release v2, I'll open a PR and perform this change.

kevinkreiser: yep of course, no problem, i wasnt sure how you like to handle that! thanks

vthiery (approved):

PR #29, opened 2021-06-15
Allocation Optimizations

kevinkreiser:
In a previous PR I noticed that we instantiated a `char[]` only to copy it into a string and then potentially copy that string further on down the chain. In that previous PR I changed the `char[]` to a fixed size string but still the string was being allocated on ever call to `send` and again potentially copied.

This PR moves that buffer to a member variable, to which we write our messages and from which we read to send bytes over the socket. When in batching mode I've also made some changes to remove as many re-allocations as possible (by reserving the batch size for any new batches that are made). I've changed the send interface of the `UDPSender` to use iterators so that we don't copy any bytes from the `StatsdClient` to the `UDPSender`.

One strange quirk that I noticed a while back that we dont handle is when a person tries to send a stat that is larger than our buffer. `snprintf` tells us when this happens but we previously didnt do anything about it. i've added a unit test for that case but annoyingly since the `UDPSender` owns the error message it was somewhat un-tidy to allow the `StatsdClient` to set this message.

### DIFF
--- a/include/cpp-statsd-client/StatsdClient.hpp
+++ b/include/cpp-statsd-client/StatsdClient.hpp
@@ -152,14 +152,13 @@ inline void StatsdClient::send(const std::string& key,
         return;
     }
 
-    // Prepare the buffer and include the sampling rate if it's not 1.f
-    int string_len = -1;
+    // Sampling rate is 1.0f, no need to specify it
+    int string_len;
     if (isFrequencyOne) {
-        // Sampling rate is 1.0f, no need to specify it
         string_len = std::snprintf(
             &m_buffer.front(), m_buffer.size(), "%s%s:%d|%s", m_prefix.c_str(), key.c_str(), value, type.c_str());
-    } else {
-        // Sampling rate is different from 1.0f, hence specify it
+    }  // Sampling rate is different from 1.0f, hence specify it
+    else {
         string_len = std::snprintf(&m_buffer.front(),
                                    m_buffer.size(),
                                    "%s%s:%d|%s|@%.2f",
@@ -168,6 +167,12 @@ inline void StatsdClient::send(const std::string& key,
                                    value,
                                    type.c_str(),
                                    frequency);
+    }
+
+    // Your stat was too large
+    if (static_cast<size_t>(string_len) > m_buffer.size()) {
+        m_sender->setErrorMessage("snprintf failed message too large");
+        return;
     }
 
     // Send the message via the UDP sender
--- a/include/cpp-statsd-client/StatsdClient.hpp
+++ b/include/cpp-statsd-client/StatsdClient.hpp
@@ -109,9 +109,11 @@ inline StatsdClient::StatsdClient(const std::string& host,
                                   const uint16_t port,
                                   const std::string& prefix,
                                   const uint64_t batchsize) noexcept
-    : m_prefix(detail::sanitizePrefix(prefix)), m_sender(new UDPSender{host, port, batchsize}), m_buffer(256, '\0') {
+    : m_prefix(detail::sanitizePrefix(prefix)), m_sender(new UDPSender{host, port, batchsize}) {
     // Initialize the random generator to be used for sampling
     seed();
+    // Avoid re-allocations by reserving a generous buffer
+    m_buffer.reserve(256);
 }
 
 inline void StatsdClient::setConfig(const std::string& host,
@@ -168,39 +170,22 @@ inline void StatsdClient::send(const std::string& key,
         return;
     }
 
-    // Sampling rate is 1.0f, no need to specify it
-    const bool needs_dot = !m_prefix.empty() && !key.empty();
-    int string_len;
-    if (isFrequencyOne) {
-        string_len = std::snprintf(&m_buffer.front(),
-                                   m_buffer.size(),
-                                   "%s%s%s:%d|%s",
-                                   m_prefix.c_str(),
-                                   needs_dot ? "." : "",
-                                   key.c_str(),
-                                   value,
-                                   type.c_str());
-    }  // Sampling rate is different from 1.0f, hence specify it
-    else {
-        string_len = std::snprintf(&m_buffer.front(),
-                                   m_buffer.size(),
-                                   "%s%s%s:%d|%s|@%.2f",
-                                   m_prefix.c_str(),
-                                   needs_dot ? "." : "",
-                                   key.c_str(),
-                                   value,
-                                   type.c_str(),
-                                   frequency);
-    }
-
-    // Your stat was too large
-    if (static_cast<size_t>(string_len) > m_buffer.size()) {
-        m_sender->setErrorMessage("snprintf failed message too large");
-        return;
+    // Format the stat message
+    m_buffer.clear();
+    m_buffer.append(m_prefix);
+    if (!m_prefix.empty() && !key.empty()) m_buffer.push_back('.');
+    m_buffer.append(key);
+    m_buffer.push_back(':');
+    m_buffer.append(std::to_string(value));
+    m_buffer.push_back('|');
+    m_buffer.append(type);
+    if (frequency < 1.f) {
+        m_buffer.append("|@0.");
+        m_buffer.append(std::to_string(static_cast<int>(frequency * 100)));
     }
 
     // Send the message via the UDP sender
-    m_sender->send(m_buffer.cbegin(), m_buffer.cbegin() + string_len);
+    m_sender->send(m_buffer);
 }
 
 inline void StatsdClient::seed(unsigned int seed) noexcept {
--- a/include/cpp-statsd-client/StatsdClient.hpp
+++ b/include/cpp-statsd-client/StatsdClient.hpp
@@ -84,13 +84,16 @@ private:
 
     //! The random number generator for handling sampling
     mutable std::mt19937 m_randomEngine;
+
+    //! The buffer string format our stats before sending them
+    mutable std::string m_buffer;
 };
 
 inline StatsdClient::StatsdClient(const std::string& host,
                                   const uint16_t port,
                                   const std::string& prefix,
                                   const uint64_t batchsize) noexcept
-    : m_prefix(prefix), m_sender(new UDPSender{host, port, batchsize}) {
+    : m_prefix(prefix), m_sender(new UDPSender{host, port, batchsize}), m_buffer(256, '\0') {
     // Initialize the random generator to be used for sampling
     seed();
 }
@@ -150,16 +153,15 @@ inline void StatsdClient::send(const std::string& key,
     }
 
     // Prepare the buffer and include the sampling rate if it's not 1.f
-    std::string buffer(256, '\0');
     int string_len = -1;
     if (isFrequencyOne) {
         // Sampling rate is 1.0f, no need to specify it
         string_len = std::snprintf(
-            &buffer.front(), buffer.size(), "%s%s:%d|%s", m_prefix.c_str(), key.c_str(), value, type.c_str());
+            &m_buffer.front(), m_buffer.size(), "%s%s:%d|%s", m_prefix.c_str(), key.c_str(), value, type.c_str());
     } else {
         // Sampling rate is different from 1.0f, hence specify it
-        string_len = std::snprintf(&buffer.front(),
-                                   buffer.size(),
+        string_len = std::snprintf(&m_buffer.front(),
+                                   m_buffer.size(),
                                    "%s%s:%d|%s|@%.2f",
                                    m_prefix.c_str(),
                                    key.c_str(),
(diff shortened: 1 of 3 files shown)


include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -84,13 +84,16 @@ class StatsdClient {
...
+    //! The buffer string format our stats before sending them
+    mutable std::string m_buffer;
 };
 
 inline StatsdClient::StatsdClient(const std::string& host,
                                   const uint16_t port,
                                   const std::string& prefix,
                                   const uint64_t batchsize) noexcept
-    : m_prefix(prefix), m_sender(new UDPSender{host, port, batchsize}) {
+    : m_prefix(prefix), m_sender(new UDPSender{host, port, batchsize}), m_buffer(256, '\0') {
```
kevinkreiser, line 96: 256 seems big enough to me but maybe sometime in the future we could make it configurable

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -149,17 +152,15 @@ inline void StatsdClient::send(const std::string& key,
...
-            &buffer.front(), buffer.size(), "%s%s:%d|%s", m_prefix.c_str(), key.c_str(), value, type.c_str());
-    } else {
-        // Sampling rate is different from 1.0f, hence specify it
-        string_len = std::snprintf(&buffer.front(),
-                                   buffer.size(),
+            &m_buffer.front(), m_buffer.size(), "%s%s:%d|%s", m_prefix.c_str(), key.c_str(), value, type.c_str());
+    }  // Sampling rate is different from 1.0f, hence specify it
+    else {
+        string_len = std::snprintf(&m_buffer.front(),
+                                   m_buffer.size(),
```
kevinkreiser, line 163: this is hard to read because i moved the comments above the control flow statements, but all i really did here is change `buffer` to `m_buffer`

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -149,17 +152,15 @@ inline void StatsdClient::send(const std::string& key,
         return;
     }
 
-    // Prepare the buffer and include the sampling rate if it's not 1.f
-    std::string buffer(256, '\0');
```
kevinkreiser, line 153: now we dont have to make a new string every time we send

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -168,12 +169,14 @@ inline void StatsdClient::send(const std::string& key,
                                    frequency);
     }
 
-    // Trim the trailing null chars
-    // TODO: perf improvement: send the len along and keep the buffer around as a member variable
-    buffer.resize(std::min(string_len, static_cast<int>(buffer.size())));
+    // Your stat was too large
+    if (static_cast<size_t>(string_len) > m_buffer.size()) {
+        m_sender->setErrorMessage("snprintf failed message too large");
```
kevinkreiser, line 174: previously we would send the malformed message to statsd, now we error and tell the caller.

@vthiery i found this a bit messy since this class doesnt own the error message. do you have any better suggestions? we could make the client a friend class of the sender, that would work too and be less verbose but im not sure which you'd prefer.

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -168,12 +169,14 @@ inline void StatsdClient::send(const std::string& key,
...
+        return;
+    }
 
     // Send the message via the UDP sender
-    m_sender->send(buffer);
+    m_sender->send(m_buffer.cbegin(), m_buffer.cbegin() + string_len);
```
kevinkreiser, line 179: we use iterater ranges to talk about bytes in the string without copying them

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -166,30 +169,39 @@ inline UDPSender::~UDPSender() {
...
     // If batching is on, accumulate messages in the queue
     if (m_batching) {
         std::unique_lock<std::mutex> batchingLock(m_batchingMutex);
+        // Either we don't have a place to batch our message or we exceeded the batch size, so make a new batch
         if (m_batchingMessageQueue.empty() || m_batchingMessageQueue.back().length() > m_batchsize) {
-            m_batchingMessageQueue.push_back(message);
-        } else {
-            m_batchingMessageQueue.rbegin()->append("\n").append(message);
+            m_batchingMessageQueue.emplace_back();
+            m_batchingMessageQueue.back().reserve(m_batchsize + 256);
```
kevinkreiser, line 180: when we need a new batch started, we now reserve enough bytes that we will not need to reallocate it again.

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -213,7 +225,7 @@ inline bool UDPSender::initialize() noexcept {
             // An error code has been returned by getaddrinfo
             close(m_socket);
             m_socket = k_invalidFd;
-            m_errorMessage = "getaddrinfo failed: error=" + std::to_string(ret) + ", msg=" + gai_strerror(ret);
+            m_errorMessage = "getaddrinfo failed: err=" + std::to_string(ret) + ", msg=" + gai_strerror(ret);
```
kevinkreiser, line 228: i tried to make all the error message look as much like each other as possible. its still not perfect but slightly better i guess

tests/testStatsdClient.cpp
```diff
@@ -31,21 +31,24 @@ void mock(StatsdServer& server, std::vector<std::string>& messages) {
     } while (server.errorMessage().empty() && !messages.back().empty());
 }
 
-// TODO: would it be ok if we change the API of the client to throw on error instead of having to check a string?
-void throwOnError(const StatsdClient& client) {
-    if (!client.errorMessage().empty()) {
-        std::cerr << client.errorMessage() << std::endl;
-        throw std::runtime_error(client.errorMessage());
+void throwOnError(const StatsdClient& client, bool expectEmpty = true, const std::string& extraMessage = "") {
```
kevinkreiser, line 34: i realized i wanted to test when i expected there not to be an error AND when i expected there to be an error. so i extended this helper to allow for testing both possiblities

tests/testStatsdClient.cpp
```diff
@@ -31,21 +31,24 @@ void mock(StatsdServer& server, std::vector<std::string>& messages) {
...
 
 void testErrorConditions() {
-    // Connect to a rubbish ip and make sure initialization failed
+    // Resolve a rubbish ip and make sure initialization failed
     StatsdClient client{"256.256.256.256", 8125, "myPrefix.", 20};
-    if (client.errorMessage().empty()) {
-        std::cerr << "Should not be able to connect to ridiculous ip" << std::endl;
-        throw std::runtime_error("Should not be able to connect to ridiculous ip");
-    }
+    throwOnError(client, false, "Should not be able to resolve a ridiculous ip");
```
kevinkreiser, line 44: so now testing for when we expect an error is way less verbose.

vthiery:
Thanks for the work and all the comments! I'll review it asap

kevinkreiser:
@vthiery in the mean time i will get the conflict resolved! thank you so much!

kevinkreiser:
@vthiery the conflicts should be resolved now, let me know what you think!

kevinkreiser:
@vthiery is there anything i should do here to make this easier for review or are you just trying to find the time for it? let me know if there is anything else i should do here!

vthiery:
> @vthiery is there anything i should do here to make this easier for review or are you just trying to find the time for it? let me know if there is anything else i should do here!

Sorry for the delay. There is nothing left to do on your side, I just need to find some time to review it properly. I'll try to do it tomorrow.

kevinkreiser:
Well I started implementing the other PR for protecting against sending the wrong stats types and I was adding support for tags and I started to question what would be the best way to support tags. This lead me down a rabbit hole of what is the best way to format a string. I was pretty shocked at the results, but checkout this gist:

https://gist.github.com/kevinkreiser/1a7c9a7cc803a51e79f76a5e0e53730f#file-string_format-cpp-L17-L20

Based on my testing it seems using string operations directly is faster than printf-like functions, which I think makes decent sense. So I'm going to refactor this PR to do that as well.

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -165,39 +170,22 @@ inline void StatsdClient::send(const std::string& key,
...
-        string_len = std::snprintf(&buffer.front(),
-                                   buffer.size(),
-                                   "%s%s%s:%d|%s|@%.2f",
-                                   m_prefix.c_str(),
-                                   needs_dot ? "." : "",
-                                   key.c_str(),
-                                   value,
-                                   type.c_str(),
-                                   frequency);
+    // Format the stat message
```
kevinkreiser, line 173: it was faster to use string operations than `snprintf`, i benchmarked it here: https://gist.github.com/kevinkreiser/1a7c9a7cc803a51e79f76a5e0e53730f#file-string_format-cpp

the other benefit is that now we dont have to worry if the string we are format is larger than the buffer we have. when for some reason there is a really large stat message, the std::string will allocate more space to fit it. in practice this should be very unlikely.

kevinkreiser:
Ok cool I've updated the code to use string operations which simplifies this PR, makes the code faster, and makes the code easier to update in my next PR :smile:

include/cpp-statsd-client/UDPSender.hpp
vthiery: We could make this function slightly clearer with something like

```cpp
inline void UDPSender::send(const std::string& message) noexcept {
    m_errorMessage.clear();
    // If batching is on, accumulate messages in the queue
    if (m_batching) {
        queueMessage(message);
        return;
    }

    sendToDaemon(message);
}

inline void UDPSender::queueMessage(const std::string& message) noexcept {
    std::unique_lock<std::mutex> batchingLock(m_batchingMutex);
    // Either we don't have a place to batch our message or we exceeded the batch size, so make a new batch
    if (m_batchingMessageQueue.empty() || m_batchingMessageQueue.back().length() > m_batchsize) {
        m_batchingMessageQueue.emplace_back();
        m_batchingMessageQueue.back().reserve(m_batchsize + 256);
    }  // When there is already a batch open we need a separator when its not empty
    else if (!m_batchingMessageQueue.back().empty()) {
        m_batchingMessageQueue.back().push_back('\n');
    }
    // Add the new message to the batch
    m_batchingMessageQueue.back().append(message);
}
```

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -165,39 +170,22 @@ inline void StatsdClient::send(const std::string& key,
...
+    m_buffer.append(m_prefix);
+    if (!m_prefix.empty() && !key.empty()) m_buffer.push_back('.');
+    m_buffer.append(key);
+    m_buffer.push_back(':');
+    m_buffer.append(std::to_string(value));
+    m_buffer.push_back('|');
+    m_buffer.append(type);
+    if (frequency < 1.f) {
+        m_buffer.append("|@0.");
+        m_buffer.append(std::to_string(static_cast<int>(frequency * 100)));
```
vthiery, line 184: Cosmetic but makes it easier to read imo
```suggestion
    m_buffer.clear();
    
    m_buffer.append(m_prefix);
    if (!m_prefix.empty() && !key.empty()) {
        m_buffer.push_back('.');
    }
    
    m_buffer.append(key);
    m_buffer.push_back(':');
    m_buffer.append(std::to_string(value));
    m_buffer.push_back('|');
    m_buffer.append(type);
    
    if (frequency < 1.f) {
        m_buffer.append("|@0.");
        m_buffer.append(std::to_string(static_cast<int>(frequency * 100)));
```

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -165,39 +170,22 @@ inline void StatsdClient::send(const std::string& key,
...
-                                   buffer.size(),
-                                   "%s%s%s:%d|%s|@%.2f",
-                                   m_prefix.c_str(),
-                                   needs_dot ? "." : "",
-                                   key.c_str(),
-                                   value,
-                                   type.c_str(),
-                                   frequency);
+    // Format the stat message
+    m_buffer.clear();
```
vthiery, line 174: Now that we use and alter the buffer, we need to remove the `const` from pretty much all the functions too

vthiery (commented):
Looks good! Only a couple of minor details to fix

include/cpp-statsd-client/StatsdClient.hpp
kevinkreiser: i think using `mutable` and keeping the methods `const` are a perfectly reasonable here because the caller doesnt percieve any change in state of the client itself. in other words the methods are all still logically const even if they may not be bitwise const. https://en.cppreference.com/w/cpp/language/cv also has a nice little blurb that puts it pretty well:

> Mutable is used to specify that the member does not affect the externally visible state of the class (as often used for mutexes, memo caches, lazy evaluation, and access instrumentation).

include/cpp-statsd-client/UDPSender.hpp
kevinkreiser: done!

include/cpp-statsd-client/StatsdClient.hpp
kevinkreiser: @vthiery let me know if you'd still rather un`const`ify these methods and remove the `mutable` keyword

vthiery: No you're right sorry for the confusion, `mutable` and keeping `const` is fine.

kevinkreiser:
ok lint should be good now, there were spaces in the extra lines between the string formating code blocks. thanks again @vthiery!

vthiery (approved):

vthiery:
Thanks for the contribution @kevinkreiser !

PR #30, opened 2021-06-22
Add Missing Metric Types and Tags Support

### LINKED ISSUE #11: Support for statsd tags
We were wondering if tags are supported by this client library and how they should be used. 
We are looking something similar to python statsd client metrics tagging feature
```
    import statsd
    c = statsd.StatsClient('localhost', 8125, prefix='foo')
    c.incr('bar', tags = {"key1": "value1", "key2":"value2"})
```
As discussed here https://stackoverflow.com/questions/49852654/sending-statsd-metrics-with-tags-from-python-client

### LINKED ISSUE #26: Add Sets and Histograms (Alias of Counters) and Guard the Metric Type Interface
This issue is based on the review comment here: https://github.com/vthiery/cpp-statsd-client/pull/19#discussion_r647179020

The idea here is that we'd like to make it impossible for the client to send messages with bogus formatting but the current API allows users to send metrics manually by specifying the metric type as a string. This opens the door for users to pass a string that is not valid and would force us to validate it and set an error message if invalid. With a small (but breaking) API change we can side step the whole issue by:

1. making the `send` method `private`
2. add support for all the valid metric types (`set` and `histogram` are missing)

To follow with semver we will need to increment the major revision though.

@vthiery In general do you think this is a worthwhile thing to do?

kevinkreiser:
As the title says! We also now protect against sending made up metric types by hiding the raw sending function.

Fixes #11 
Fixes #26

### DIFF
--- a/include/cpp-statsd-client/StatsdClient.hpp
+++ b/include/cpp-statsd-client/StatsdClient.hpp
@@ -87,13 +87,6 @@ public:
                 float frequency = 1.0f,
                 const std::vector<std::string>& tags = {}) const noexcept;
 
-    //! Records an arbitrary unit for a key, at a given frequency. An alias for timing
-    //! suited to use cases whose units do not involve time
-    void histogram(const std::string& key,
-                   const unsigned int value,
-                   float frequency = 1.0f,
-                   const std::vector<std::string>& tags = {}) const noexcept;
-
     //! Records a count of unique occurrences for a key, at a given frequency
     void set(const std::string& key,
              const unsigned int sum,
@@ -145,7 +138,6 @@ std::string sanitizePrefix(std::string prefix) {
 constexpr char METRIC_TYPE_COUNT[] = "c";
 constexpr char METRIC_TYPE_GUAGE[] = "g";
 constexpr char METRIC_TYPE_TIMING[] = "ms";
-constexpr char METRIC_TYPE_HISTOGRAM[] = "h";
 constexpr char METRIC_TYPE_SET[] = "s";
 }  // namespace detail
 
@@ -203,13 +195,6 @@ inline void StatsdClient::timing(const std::string& key,
                                  float frequency,
                                  const std::vector<std::string>& tags) const noexcept {
     return send(key, ms, detail::METRIC_TYPE_TIMING, frequency, tags);
-}
-
-inline void StatsdClient::histogram(const std::string& key,
-                                    const unsigned int value,
-                                    float frequency,
-                                    const std::vector<std::string>& tags) const noexcept {
-    return send(key, value, detail::METRIC_TYPE_HISTOGRAM, frequency, tags);
 }
 
 inline void StatsdClient::set(const std::string& key,
--- a/include/cpp-statsd-client/StatsdClient.hpp
+++ b/include/cpp-statsd-client/StatsdClient.hpp
@@ -73,11 +73,24 @@ public:
     //! Records a timing for a key, at a given frequency
     void timing(const std::string& key, const unsigned int ms, float frequency = 1.0f) const noexcept;
 
-    //! Send a value for a key, according to its type, at a given frequency
-    void send(const std::string& key, const int value, const std::string& type, float frequency = 1.0f) const noexcept;
+    //! Records an arbitrary unit for a key, at a given frequency. An alias for timing
+    //! suited to use cases whose units do not involve time
+    void histogram(const std::string& key, const unsigned int value, float frequency = 1.0f) const noexcept;
+
+    //! Records a count of unique occurrences for a key, at a given frequency
+    void set(const std::string& key, const unsigned int sum, float frequency = 1.0f) const noexcept;
 
     //! Seed the RNG that controls sampling
     void seed(unsigned int seed = std::random_device()()) noexcept;
+
+    //!@}
+
+private:
+    // @name Private methods
+    // @{
+
+    //! Send a value for a key, according to its type, at a given frequency
+    void send(const std::string& key, const int value, const char* type, float frequency = 1.0f) const noexcept;
 
     //!@}
 
@@ -100,6 +113,13 @@ std::string sanitizePrefix(std::string prefix) {
     }
     return prefix;
 }
+
+// All supported metric types
+constexpr char METRIC_TYPE_COUNT[] = "c";
+constexpr char METRIC_TYPE_GUAGE[] = "g";
+constexpr char METRIC_TYPE_TIMING[] = "ms";
+constexpr char METRIC_TYPE_HISTOGRAM[] = "h";
+constexpr char METRIC_TYPE_SET[] = "s";
 }  // namespace detail
 
 inline StatsdClient::StatsdClient(const std::string& host,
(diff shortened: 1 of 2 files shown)


include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -59,28 +60,64 @@ class StatsdClient {
     const std::string& errorMessage() const noexcept;
 
     //! Increments the key, at a given frequency rate
-    void increment(const std::string& key, float frequency = 1.0f) const noexcept;
+    void increment(const std::string& key,
```
kevinkreiser, line 63: since i added tags, all of the method signatures changed. its hard to check over but thats the only diff to the method signatures

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -59,28 +60,64 @@ class StatsdClient {
...
     void seed(unsigned int seed = std::random_device()()) noexcept;
 
     //!@}
 
+private:
+    // @name Private methods
+    // @{
+
+    //! Send a value for a key, according to its type, at a given frequency
+    void send(const std::string& key,
```
kevinkreiser, line 113: we move send to private so that you cant accidentally send a bad message

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -103,6 +140,13 @@ std::string sanitizePrefix(std::string prefix) {
     }
     return prefix;
 }
+
+// All supported metric types
+constexpr char METRIC_TYPE_COUNT[] = "c";
+constexpr char METRIC_TYPE_GUAGE[] = "g";
+constexpr char METRIC_TYPE_TIMING[] = "ms";
+constexpr char METRIC_TYPE_HISTOGRAM[] = "h";
+constexpr char METRIC_TYPE_SET[] = "s";
```
kevinkreiser, line 149: here we enumerate all of the possible metric types rather than hardcode them in the various methods i thougt it was nicer to have them all together hidden in the detail namespace

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -189,6 +259,15 @@ inline void StatsdClient::send(const std::string& key,
         m_buffer.append(std::to_string(static_cast<int>(frequency * 100)));
     }
 
+    if (!tags.empty()) {
```
kevinkreiser, line 262: here we add tags support

tests/testStatsdClient.cpp
```diff
@@ -48,24 +48,24 @@ void testReconfigure() {
...
 
     client.setConfig("localhost", 8125, "");
-    client.send("third.baz", 1, "c", 1.f);
+    client.increment("third.baz");
     if (server.receive() != "third.baz:1|c") {
         throw std::runtime_error("Incorrect stat received");
     }
 
-    client.send("", 1, "c", 1.f);
+    client.increment("");
```
kevinkreiser, line 68: send is now private

tests/testStatsdClient.cpp
```diff
@@ -118,14 +118,29 @@ void testSendRecv(uint64_t batchSize) {
...
+
+        // Gauge but with tags
+        client.gauge("grabe", 333, 1.f, {"liegt", "im", "weste"});
+        throwOnError(client);
+        expected.emplace_back("sendRecv.grabe:333|g|#liegt,im,weste");
+
+        // All the things
+        client.count("foo", -42, .9f, {"bar", "baz"});
+        throwOnError(client);
+        expected.emplace_back("sendRecv.foo:-42|c|@0.90|#bar,baz");
```
kevinkreiser, line 139: these are all the new metrics and tags we can send

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -103,6 +140,13 @@ std::string sanitizePrefix(std::string prefix) {
     }
     return prefix;
 }
+
+// All supported metric types
+constexpr char METRIC_TYPE_COUNT[] = "c";
+constexpr char METRIC_TYPE_GUAGE[] = "g";
+constexpr char METRIC_TYPE_TIMING[] = "ms";
+constexpr char METRIC_TYPE_HISTOGRAM[] = "h";
```
vthiery, line 148: Do you have a link that details how histogram should be used? I didn't find them in the [metric types doc](https://github.com/statsd/statsd/blob/master/docs/metric_types.md) (that we should probably link in the README somewhere actually) and could only find a mention in [gostatsd](https://github.com/atlassian/gostatsd#timer-histograms-experimental-feature).

kevinkreiser: i forget where i first read about it but looking at the code for the 100th time it seems like it would get stored as a counter, which doesnt seem right: https://github.com/statsd/statsd/blob/9cf77d87855bcb69a8663135f59aa23825db9797/stats.js#L317-L322

so ill remove it

kevinkreiser: done

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -59,28 +60,57 @@ class StatsdClient {
     const std::string& errorMessage() const noexcept;
 
     //! Increments the key, at a given frequency rate
-    void increment(const std::string& key, float frequency = 1.0f) const noexcept;
+    void increment(const std::string& key,
+                   float frequency = 1.0f,
+                   const std::vector<std::string>& tags = {}) const noexcept;
```
vthiery, line 65: `std::vector` is fine but do you know if there is something preventing us from using a variadic argument instead?
Could that be easier to use somehow?

tests/testStatsdClient.cpp
```diff
@@ -118,14 +118,24 @@ void testSendRecv(uint64_t batchSize) {
...
-        // Send a metric explicitly
-        client.send("tutu", 241, "c", 2.0f);
+        // Send a set with 1227 total uniques
+        client.set("tutu", 1227, 2.0f);
         throwOnError(client);
-        expected.emplace_back("sendRecv.tutu:241|c");
+        expected.emplace_back("sendRecv.tutu:1227|s");
+
+        // Gauge but with tags
+        client.gauge("dr.rösti.grabe", 333, 1.f, {"liegt", "im", "weste"});
```
vthiery, line 127: :smile:

include/cpp-statsd-client/StatsdClient.hpp
kevinkreiser: the main reason is that `frequency` has a default. so the compiler says, "why don't your var args have a default too"

kevinkreiser: the other option is to use variadic templates but that is much messier than looping over a vector i think. should i give that a try though just in case, ive not really used them before so it might be fun to try it out!

vthiery: No need to dig too much into it. As I said, `std::vector` is fine, so let's go with that :+1:

tests/testStatsdClient.cpp
kevinkreiser: i thought it would be a funny easter egg in the tests since my ancestors come from the bodensee area near bischofszell. we still speak an old alemannisch dialect here but i visit switzerland as often as i can :smile:

vthiery (approved):

vthiery:
Thanks for adding this long-awaited feature!

PR #31, opened 2021-06-24
More Control Over How/When to Send to the Daemon

### LINKED ISSUE #20: Feature: Extended Stats Batching Controls
At the moment we have 2 modes for `StatsdClient`:

1. a simple mode in which every stat is sent immediately in a blocking fashion
2. a batching mode in which every stat is sent to a queue, from which a background thread will, on a secondly basis, send the queued stats from the last round. there is also a batch size which allows making use of the [multi metric feature](https://github.com/statsd/statsd/blob/master/docs/metric_types.md#multi-metric-packets)

I propose to make this interface a little more customizable by allowing each of these pieces to be configured individually. The specific pieces I'm talking about are:

- **when to send?**
  - immediately
  - on some timer in a background thread
  - on some number of stats
  - manually when you tell it to
- **what batch size to use?**
  - this isnt useful when immediately sending but when using a background thread or when deciding when to send on your own it is useful

Basically I'm hoping to refactor the configuration to allow more control by the end application on how and when the bytes get shipped over the socket.

In my specific case I have an application which does a bunch of heavy processing in a tight loop, when its done it returns the result to a client who requested it and then does some cleanup. This cleanup is done outside of the request loop, ie it returns the result before it blocks on cleaning up so that the requester isnt waiting on cleanup. This cleanup makes an idea spot for me to fire off any stats i've collected during the processing of that request. I could make use of the batchign as-is but i dont want:

1. more threads than i already have in this application
2. any synchronization at all within the hot loop of my program

For this reason it would be really useful to add a mode in which I can set a batch size to whatever maximum makes sense on my network. Have a queue to which my hot loop adds stats as it runs. And then have a method on the `StatsdClient` that allows me to flush that queue in my cleanup after my algorithm has run.

It probably goes without saying but I'm happy to implement this feature so please feel free to assign me to it unless it sounds like something you'd rather not have in the library.

kevinkreiser:
fixes #20 

The brief idea here is that we would like more explicit control over how and when to send a batch of stats to the statsd process. This PR adds one more control, specifically the time interval at which stats batches are sent. This API change is not breaking in terms of a major version as the default functionality is maintained but a new parameter is added to the configuration which allows this finer grained control. In addition to this finer control we also allow manual flushing (ie on-demand flushing of batches of stats in the queue). This allows a calling application to queue/batch stats as much as they want and then decide on thier own when is a good time to send those to statsd.

I've updated the documentation in the code to spell this out but this could probably also be done in the README, ~so I'll add that too~ I've added that.

~I'll put this in draft for now as I didnt write the unit tests yet. Once I have the tests and the README updated I'll take it out of draft~ this has all been done.

@vthiery let me know what your opinion is on this extra level of control. I also thought maybe I should add an accessor to return the queue depth as it might inform the callers decision about when to flush the stats if they are doing it manually.

### DIFF
--- a/include/cpp-statsd-client/StatsdClient.hpp
+++ b/include/cpp-statsd-client/StatsdClient.hpp
@@ -87,13 +87,6 @@ public:
                 float frequency = 1.0f,
                 const std::vector<std::string>& tags = {}) const noexcept;
 
-    //! Records an arbitrary unit for a key, at a given frequency. An alias for timing
-    //! suited to use cases whose units do not involve time
-    void histogram(const std::string& key,
-                   const unsigned int value,
-                   float frequency = 1.0f,
-                   const std::vector<std::string>& tags = {}) const noexcept;
-
     //! Records a count of unique occurrences for a key, at a given frequency
     void set(const std::string& key,
              const unsigned int sum,
@@ -145,7 +138,6 @@ std::string sanitizePrefix(std::string prefix) {
 constexpr char METRIC_TYPE_COUNT[] = "c";
 constexpr char METRIC_TYPE_GUAGE[] = "g";
 constexpr char METRIC_TYPE_TIMING[] = "ms";
-constexpr char METRIC_TYPE_HISTOGRAM[] = "h";
 constexpr char METRIC_TYPE_SET[] = "s";
 }  // namespace detail
 
@@ -203,13 +195,6 @@ inline void StatsdClient::timing(const std::string& key,
                                  float frequency,
                                  const std::vector<std::string>& tags) const noexcept {
     return send(key, ms, detail::METRIC_TYPE_TIMING, frequency, tags);
-}
-
-inline void StatsdClient::histogram(const std::string& key,
-                                    const unsigned int value,
-                                    float frequency,
-                                    const std::vector<std::string>& tags) const noexcept {
-    return send(key, value, detail::METRIC_TYPE_HISTOGRAM, frequency, tags);
 }
 
 inline void StatsdClient::set(const std::string& key,
--- a/include/cpp-statsd-client/StatsdClient.hpp
+++ b/include/cpp-statsd-client/StatsdClient.hpp
@@ -24,11 +24,33 @@ namespace Statsd {
  * key, therefore you should neither append one to the prefix
  * nor prepend one to the key
  *
- * The sampling frequency can be input, as well as the
- * batching size. The latter is the optional limit in
- * number of bytes a batch of stats can be before it is
- * new states are added to a fresh batch. A value of 0
- * means batching is disabled.
+ * The sampling frequency is specified per call and uses a
+ * random number generator to determine whether or not the stat
+ * will be recorded this time or not.
+ *
+ * The top level configuration includes 2 optional parameters
+ * that determine how the stats are delivered to statsd. These
+ * parameters are the batching size and the send interval.
+ *
+ * The batching size controls the number of bytes to send
+ * in each UDP datagram to statsd. This is not a hard limit as
+ * we continue appending to a batch of stats until the limit
+ * has been reached or surpassed. When this occurs we add the
+ * batch to a queue and create a new batch to appended to. A
+ * value of 0 for the batching size will disable batching such
+ * that each stat will be sent to the daemon individually.
+ *
+ * The send interval controls the rate at which queued batches
+ * of stats will be sent to statsd. If batching is disabled,
+ * this value is ignored and each individual stat is sent to
+ * statsd immediately in a blocking fashion. If batching is
+ * enabled (ie. non-zero) then the send interval is the number
+ * of milliseconds to wait before flushing the queue of
+ * batched stats messages to the daemon. This is done in a non-
+ * blocking fashion via a background thread. If the send
+ * interval is 0 then the stats messages are appended to a
+ * queue until the caller manually flushes the queue via the
+ * flush method.
  *
  */
 class StatsdClient {
@@ -40,7 +62,8 @@ public:
     StatsdClient(const std::string& host,
                  const uint16_t port,
                  const std::string& prefix,
-                 const uint64_t batchsize = 0) noexcept;
+                 const uint64_t batchsize = 0,
+                 const uint64_t sendInterval = 1000) noexcept;
 
     StatsdClient(const StatsdClient&) = delete;
     StatsdClient& operator=(const StatsdClient&) = delete;
@@ -54,7 +77,8 @@ public:
     void setConfig(const std::string& host,
                    const uint16_t port,
                    const std::string& prefix,
-                   const uint64_t batchsize = 0) noexcept;
+                   const uint64_t batchsize = 0,
+                   const uint64_t sendInterval = 1000) noexcept;
 
     //! Returns the error message as an std::string
     const std::string& errorMessage() const noexcept;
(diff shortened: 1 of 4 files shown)


include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -40,7 +62,8 @@ class StatsdClient {
     StatsdClient(const std::string& host,
                  const uint16_t port,
                  const std::string& prefix,
-                 const uint64_t batchsize = 0) noexcept;
+                 const uint64_t batchsize = 0,
+                 const uint64_t sendInterval = 1000) noexcept;
```
kevinkreiser, line 66: so now we allow the caller to configure the interval at which queued stats batches are sent. this implies that you are batching. here we default to the previously hardcoded value of 1 second. if you set the value to 0 then automatic sending via the background thread is disabled and the caller must manually flush the batches themselves.

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -93,14 +99,14 @@ class UDPSender final {
...
-    bool m_batching{false};
-
     //! The batching size
     uint64_t m_batchsize;
 
+    //! The sending frequency in milliseconds
+    uint64_t m_sendInterval;
+
     //! The queue batching the messages
-    mutable std::deque<std::string> m_batchingMessageQueue;
```
kevinkreiser, line 103: i dont know why this was marked mutable, its in a non-const method with imho is properly non-const because queueing is a change to internal state that the caller should be aware of

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -117,24 +123,20 @@ class UDPSender final {
...
     // If batching is on, use a dedicated thread to send after the wait time is reached
-    if (batchsize != 0) {
-        // Thread' sleep duration between batches
-        // TODO: allow to input this
-        constexpr unsigned int batchingWait{1000U};
-
-        m_batching = true;
-        m_batchsize = batchsize;
-
+    if (m_batchsize != 0 && m_sendInterval > 0) {
```
kevinkreiser, line 137: we only have an automatic sending of the queued batches if the sender is configured to send at a non zero interval

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -161,27 +163,33 @@ inline UDPSender::~UDPSender() {
         return;
     }
 
-    if (m_batching) {
+    // If we're running a background thread tell it to stop
+    if (m_batchingThread.joinable()) {
```
kevinkreiser, line 167: `m_batching` really just meant are we running a background thread, which is easily checkable with the `joinable` function

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -161,27 +163,33 @@ inline UDPSender::~UDPSender() {
...
+    // Cleanup the socket
     close(m_socket);
 }
 
 inline void UDPSender::send(const std::string& message) noexcept {
     m_errorMessage.clear();
+
     // If batching is on, accumulate messages in the queue
-    if (m_batching) {
+    if (m_batchsize > 0) {
```
kevinkreiser, line 180: if we are batching we need to enqueue the stat rather than send it right away, regardless of whether that is for the background thread or for the caller to manually flush it later

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -161,27 +163,33 @@ inline UDPSender::~UDPSender() {
...
 
+    // Or send it right now
     sendToDaemon(message);
 }
 
 inline void UDPSender::queueMessage(const std::string& message) noexcept {
-    std::unique_lock<std::mutex> batchingLock(m_batchingMutex);
+    // We aquire a lock but only if we actually need to (ie there is a thread also accessing the queue)
+    auto batchingLock =
+        m_batchingThread.joinable() ? std::unique_lock<std::mutex>(m_batchingMutex) : std::unique_lock<std::mutex>();
```
kevinkreiser, line 192: we only actually lock if we need to synchronize with the thread, otherwise we get an empty lock which does nothing. this is to make sure that the manual flushing mode is lock-free

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -255,6 +263,17 @@ inline bool UDPSender::initialized() const noexcept {
     return m_socket != k_invalidFd;
 }
 
+inline void UDPSender::flush() noexcept {
+    // We aquire a lock but only if we actually need to (ie there is a thread also accessing the queue)
+    auto batchingLock =
+        m_batchingThread.joinable() ? std::unique_lock<std::mutex>(m_batchingMutex) : std::unique_lock<std::mutex>();
```
kevinkreiser, line 269: technically, even if you are running with a background thread you could call `flush` and it would still work but you need to  lock first to do that. another option would be to just return if you are using the thread to send to the daemon. @vthiery do you have a preference? basically the other option would be to change this to:

```suggestion
    // The background thread takes care of sending to the daemon
    if (m_batchingThread.joinable()) {
        return;
    }
```

tests/testStatsdClient.cpp
```diff
@@ -124,9 +124,9 @@ void testSendRecv(uint64_t batchSize) {
         expected.emplace_back("sendRecv.tutu:1227|s");
 
         // Gauge but with tags
-        client.gauge("dr.rösti.grabe", 333, 1.f, {"liegt", "im", "weste"});
+        client.gauge("dr.röstigrabe", 333, 1.f, {"liegt", "im", "weste"});
```
kevinkreiser, line 127: i missed this in the last PR but this is a compound noun :smile:

tests/testStatsdClient.cpp
```diff
@@ -162,9 +167,11 @@ int main() {
     // reconfiguring how you are sending
     testReconfigure();
     // no batching
-    testSendRecv(0);
+    testSendRecv(0, 0);
     // background batching
-    testSendRecv(4);
+    testSendRecv(32, 1000);
+    // manual flushing of batches
+    testSendRecv(16, 0);
```
kevinkreiser, line 174: this is the new test case for manual flushing

kevinkreiser:
@vthiery is there anything else i can do to make this easier to review? let me know if it makes sense

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -161,27 +163,33 @@ inline UDPSender::~UDPSender() {
...
         return;
     }
 
+    // Or send it right now
     sendToDaemon(message);
 }
 
 inline void UDPSender::queueMessage(const std::string& message) noexcept {
-    std::unique_lock<std::mutex> batchingLock(m_batchingMutex);
+    // We aquire a lock but only if we actually need to (ie there is a thread also accessing the queue)
```
vthiery, line 190:
```suggestion
    // We aquire a lock but only if we actually need to (i.e. there is a thread also accessing the queue)
```

include/cpp-statsd-client/UDPSender.hpp
vthiery: This is fine as it is :+1:

vthiery (approved):
LGTM, only a cosmetic change and we can merge :+1:
Thanks!

vthiery (approved):

PR #32, opened 2021-06-25
Add Code Coverage to CI

kevinkreiser:
CI should fail if code coverage falls below 85%. Current coverage is 89.1% and most of the missing lines are error handling for errors that are hard to make happen, especially udp socket or sending errors as well as lines in the test that cant be covered unless the code breaks.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -22,7 +22,9 @@ jobs:
         run: |
           make coverage
           lines=$(lcov --summary coverage.info | grep -F lines | awk '{print $2}' | sed -e "s/%//g")
-          if (( $(echo "${lines} < 85.0" | bc -l) )); then
-              echo "Line coverage dropped below 85% to ${lines}%"
+          if (( $(echo "${lines} < ${COVERAGE_THRESHOLD}" | bc -l) )); then
+              echo "Line coverage dropped below ${COVERAGE_THRESHOLD}% to ${lines}%"
               exit 1
           fi
+        env:
+          COVERAGE_THRESHOLD: 85.0
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,28 @@
+name: Test
+
+on: [push, pull_request]
+jobs:
+  test:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: dependencies
+        shell: bash
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y -qq make cmake gcc g++ lcov
+      - name: build
+        shell: bash
+        run: |
+          export LD_LIBRARY_PATH=.:$(cat /etc/ld.so.conf.d/* | grep -vF "#" | tr "\\n" ":" | sed -e "s/:$//g")
+          cmake . -DCMAKE_BUILD_TYPE=Debug -DENABLE_COVERAGE
+          make all -j$(nproc)
+      - name: coverage
+        shell: bash
+        run: |
+          make coverage
+          lines=$(lcov --summary coverage.info | grep -F lines | awk '{print $2}' | sed -e "s/%//g")
+          if (( $(echo "${lines} < 85.0" | bc -l) )); then
+              echo "Line coverage dropped below 85% to ${lines}%"
+              exit 1
+          fi
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -2,7 +2,7 @@ name: Coverage
 
 on: [push, pull_request]
 jobs:
-  test:
+  coverage:
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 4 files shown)


cmake/CodeCoverage.cmake
```diff
@@ -0,0 +1,708 @@
+# Copyright (c) 2012 - 2017, Lars Bilke
```
kevinkreiser, line 1: this is a nice helper cmake module that makes adding coverage to our project easy. it looks to be 3 clause bsd. i dont think that is a problem but if you'd rather not add this, i can do the coverage configuration and cmake targets manually. just let me know

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,28 @@
...
+        run: |
+          export LD_LIBRARY_PATH=.:$(cat /etc/ld.so.conf.d/* | grep -vF "#" | tr "\\n" ":" | sed -e "s/:$//g")
+          cmake . -DCMAKE_BUILD_TYPE=Debug -DENABLE_COVERAGE=On
+          make all -j$(nproc)
+      - name: coverage
+        shell: bash
+        run: |
+          make coverage
+          lines=$(lcov --summary coverage.info | grep -F lines | awk '{print $2}' | sed -e "s/%//g")
+          if (( $(echo "${lines} < 85.0" | bc -l) )); then
```
kevinkreiser, line 25: we require that the project stays above 85% of all lines covered by the test. let me know if this sounds ok to you or not. currently we are at 89.1% and i think with a little more tedious testing we could get to 95%.

cmake/CodeCoverage.cmake
vthiery: I'm not a license expert but it seems that the only different between MIT and this one is 
```
# 3. Neither the name of the copyright holder nor the names of its contributors
#    may be used to endorse or promote products derived from this software without
#    specific prior written permission.
```
so it should be ok.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,28 @@
...
+jobs:
+  coverage:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: dependencies
+        shell: bash
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y -qq make cmake gcc g++ lcov
```
vthiery, line 13: Running `act -j coverage`, I got
```
| Overall coverage rate:
|   lines......: 89.1% (229 of 257 lines)
|   functions..: 100.0% (32 of 32 functions)
| Lcov code coverage info report saved in coverage.info.
| Open ./coverage/index.html in your browser to view the coverage report.
| [100%] Built target coverage
| /github/workflow/3: line 3: bc: command not found
[Coverage/coverage]   ✅  Success - coverage
```
so
```suggestion
          sudo apt-get install -y -qq make cmake gcc g++ lcov bc
```

.github/workflows/coverage.yml
vthiery: 85% is fine. We should probably have it standing out a bit more using an environment variable:
```
      - name: coverage
        shell: bash
        run: |
          make coverage
          lines=$(lcov --summary coverage.info | grep -F lines | awk '{print $2}' | sed -e "s/%//g")
          if (( $(echo "${lines} < $COVERAGE_THRESHOLD" | bc -l) )); then
              echo "Line coverage dropped below $COVERAGE_THRESHOLD% to ${lines}%"
              exit 1
          fi
        env:
          COVERAGE_THRESHOLD: 85.0
```

vthiery: Let's also add a target in the Makefile:
```
BUILD_MODE ?= Release
ENABLE_COVERAGE ?= On

build: clean
	@echo "Build in ${BUILD_MODE} mode"
	mkdir -p bin/${BUILD_MODE}
	@cd bin/${BUILD_MODE}; cmake ../../ -DCMAKE_BUILD_TYPE=${BUILD_MODE} -DENABLE_COVERAGE=${ENABLE_COVERAGE}
	@cd bin/${BUILD_MODE}; make

test: build
	@cd bin/${BUILD_MODE}; make test

coverage: build
	@cd bin/${BUILD_MODE}; make coverage

install: build
	@cd bin/${BUILD_MODE}; make install

clean:
	@rm -rf bin
```

kevinkreiser:
@vthiery i think i've added all the suggestions!

Makefile
```diff
@@ -1,6 +1,7 @@
 # simple makefile to build, test and clean
 
 BUILD_MODE ?= Release
+ENABLE_COVERAGE ?= On
```
vthiery, line 4: You also need it line 9 no?
```
@cd bin/${BUILD_MODE}; cmake ../../ -DCMAKE_BUILD_TYPE=${BUILD_MODE}  -DENABLE_COVERAGE=${ENABLE_COVERAGE}
```

vthiery (approved):

PR #33, opened 2021-06-25
try a windows build

kevinkreiser:
this is a work in progress. i'm trying to add CI step that builds the library and tests it on windows. with @rbsheth 's help we've updated the library to have the right `ifdefs` for windows but im still messing around with getting the build to work on CI.

### DIFF
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -11,7 +11,7 @@ jobs:
           choco install cmake
       - name: build
         run: |
-          cmake -H . -B build -G "Visual Studio 15 2017" -A x64
+          cmake -H. -Bbuild -G "Visual Studio 15 2017" -A x64
           cmake --build build --target ALL_BUILD --config Release
       - name: test
         run: |
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -15,4 +15,4 @@ jobs:
           cmake --build build --target ALL_BUILD --config Release
       - name: test
         run: |
-          ctest --output-on-failure --test-dir build --verbose #cmake --build build --target RUN_TESTS --config Release
+          cmake --build build --target RUN_TESTS --config Release
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -15,4 +15,4 @@ jobs:
           cmake --build build --target ALL_BUILD --config Release
       - name: test
         run: |
-          build\testStatsdClient #cmake --build build --target RUN_TESTS --config Release
+          ctest --test-dir build --verbose #cmake --build build --target RUN_TESTS --config Release
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -10,11 +10,11 @@ jobs:
         run: |
           choco install make cmake
       - name: build
-        shell: bash
         run: |
-          cmake .
+          mkdir build
+          cd build
+          cmake .. -DCOVERAGE_ENABLED=Off
           make
       - name: test
-        shell: bash
         run: |
-          make test
+          make -C test
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -12,9 +12,9 @@ jobs:
       - name: build
         shell: bash
         run: |
-          cmake . -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_SANITIZERS=On
-          make
+          cmake .
+          make all
       - name: test
         shell: bash
         run: |
-          make check
+          make test
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -8,7 +8,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: dependencies
         run: |
-          choco install make cmake
+          choco install cmake
       - name: build
         run: |
           cmake -H . -B build -G "Visual Studio 15 2017" -A x64
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -13,8 +13,8 @@ jobs:
         shell: bash
         run: |
           cmake . -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_SANITIZERS=On
-          make all
+          make
       - name: test
         shell: bash
         run: |
-          make test
+          make check
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -13,7 +13,7 @@ jobs:
         run: |
           mkdir build
           cd build
-          cmake .. -DCOVERAGE_ENABLED=Off
+          cmake ..
           make
       - name: test
         run: |
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -15,4 +15,4 @@ jobs:
           cmake --build build --target ALL_BUILD --config Release
       - name: test
         run: |
-          ctest --test-dir build --verbose #cmake --build build --target RUN_TESTS --config Release
+          ctest --output-on-failure --test-dir build --verbose #cmake --build build --target RUN_TESTS --config Release
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -11,10 +11,8 @@ jobs:
           choco install make cmake
       - name: build
         run: |
-          mkdir build
-          cd build
-          cmake ..
-          make
+          cmake -H . -B build -G "Visual Studio 15 2017" -A x64
+          cmake --build build --target ALL_BUILD --config Release
       - name: test
         run: |
           make -C test
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -15,4 +15,4 @@ jobs:
           cmake --build build --target ALL_BUILD --config Release
       - name: test
         run: |
-          make -C test
+          cmake --build build --target RUN_TESTS --config Release
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -11,7 +11,7 @@ jobs:
           choco install cmake
       - name: build
         run: |
-          cmake -H. -Bbuild -G "Visual Studio 15 2017" -A x64
+          cmake -S . -B build -G "Visual Studio 15 2017" -A x64
           cmake --build build --target ALL_BUILD --config Release
       - name: test
         run: |
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -1,0 +1,20 @@
+name: Windows
+
+on: [push, pull_request]
+jobs:
+  windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: dependencies
+        run: |
+          choco install make cmake
+      - name: build
+        shell: bash
+        run: |
+          cmake . -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_SANITIZERS=On
+          make all
+      - name: test
+        shell: bash
+        run: |
+          make test
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -15,4 +15,4 @@ jobs:
           cmake --build build --target ALL_BUILD --config Release
       - name: test
         run: |
-          cmake --build build --target RUN_TESTS --config Release
+          build\testStatsdClient #cmake --build build --target RUN_TESTS --config Release
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -11,7 +11,7 @@ jobs:
           choco install cmake
       - name: build
         run: |
-          cmake -S . -B build -G "Visual Studio 15 2017" -A x64
+          cmake -S . -B build -G "Visual Studio 16 2019" -A x64
           cmake --build build --target ALL_BUILD --config Release
       - name: test
         run: |
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,13 +8,13 @@ project(cpp-statsd-client
 
 option(BUILD_TESTS "Build tests" ON)
 option(HUNTER_ENABLED "Use the Hunter package manager" OFF)
+option(ENABLE_COVERAGE "Build with coverage instrumentalisation" OFF)
 
 if(NOT HUNTER_ENABLED)
   set(CMAKE_CXX_STANDARD 11)
   set(CMAKE_CXX_EXTENSIONS OFF)
 endif()
 
-option(ENABLE_COVERAGE "Build with coverage instrumentalisation" OFF)
 include(GNUInstallDirs)
 include(CMakePackageConfigHelpers)
 find_package(Threads)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,14 +6,14 @@ project(cpp-statsd-client
         DESCRIPTION "A header-only StatsD client implemented in C++"
         HOMEPAGE_URL "https://github.com/vthiery/cpp-statsd-client")
 
-option(ENABLE_TESTS "Build tests" ON)
-option(HUNTER_ENABLED "Use the Hunter package manager" OFF)
-option(ENABLE_COVERAGE "Build with coverage instrumentalisation" OFF)
-
-if(NOT HUNTER_ENABLED)
-  set(CMAKE_CXX_STANDARD 11)
-  set(CMAKE_CXX_EXTENSIONS OFF)
+option(CPP_STATSD_STANDALONE "Allows configuration of targets for verifying library functionality" ON)
+if(${CPP_STATSD_STANDALONE})
+  option(ENABLE_TESTS "Build tests" ON)
+  option(ENABLE_COVERAGE "Build with coverage instrumentalisation" OFF)
 endif()
+
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_EXTENSIONS OFF)
 
 include(GNUInstallDirs)
 include(CMakePackageConfigHelpers)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,8 +5,14 @@ project(cpp-statsd-client
         LANGUAGES CXX
         DESCRIPTION "A header-only StatsD client implemented in C++"
         HOMEPAGE_URL "https://github.com/vthiery/cpp-statsd-client")
-set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_CXX_EXTENSIONS OFF)
+
+option(BUILD_TESTS "Build tests" OFF)
+option(HUNTER_ENABLED "Use the Hunter package manager" OFF)
+
+if(NOT HUNTER_ENABLED)
+  set(CMAKE_CXX_STANDARD 11)
+  set(CMAKE_CXX_EXTENSIONS OFF)
+endif()
 include(GNUInstallDirs)
 include(CMakePackageConfigHelpers)
 find_package(Threads)
@@ -18,6 +24,9 @@ target_include_directories(
   INTERFACE $<BUILD_INTERFACE:${${PROJECT_NAME}_SOURCE_DIR}/include>
             $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 target_link_libraries(${PROJECT_NAME} INTERFACE Threads::Threads)
+if(WIN32)
+  target_link_libraries(${PROJECT_NAME} INTERFACE ws2_32)
+endif()
 
 # The installation and pkg-config-like cmake config
 install(TARGETS ${PROJECT_NAME}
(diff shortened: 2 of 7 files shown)


rbsheth:
@kevinkreiser You might like #34 😄

kevinkreiser:
@rbsheth looks like you are further along than i. i was starting with the CI bits while you seem to be starting off with the actual code/build changes. do you plan on adding a github action to CI?

rbsheth:
@kevinkreiser No, these are changes we use internally that have been tested to work cross-platform. You can build on my PR with a Github Action to ensure no breakage?

kevinkreiser:
@rbsheth ive cherry picked your commits into this pr. with any luck i can get CI working for windows in the next little while

vthiery:
If you don't mind, I'd prefer to first add the support for Windows and then revisit the topic of Hunter (I have no idea how this manager works neither how it should be configured)

rbsheth:
@vthiery The impact on non-Hunter users is zero since I've disabled it by default. https://github.com/vthiery/cpp-statsd-client/pull/33/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR10

The argument is only needed since Hunter manages the C++ standard/other flags across all compile units.

.github/workflows/win.yml
```diff
@@ -0,0 +1,20 @@
...
+      - uses: actions/checkout@v2
+      - name: dependencies
+        run: |
+          choco install make cmake
+      - name: build
+        run: |
+          mkdir build
+          cd build
+          cmake ..
+          make
```
rbsheth, line 17:
```suggestion
          cmake -H . -B build -G "Visual Studio 15 2017" -A x64
          cmake --build build --target ALL_BUILD --config Release
```

.github/workflows/win.yml
```diff
@@ -0,0 +1,20 @@
...
+      - name: test
+        run: |
+          make -C test
```
rbsheth, line 20:
```suggestion
          cmake --build build --target RUN_TESTS --config Release
```

.github/workflows/win.yml
```diff
@@ -0,0 +1,20 @@
...
+
+on: [push, pull_request]
+jobs:
+  windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: dependencies
+        run: |
+          choco install make cmake
```
rbsheth, line 11:
```suggestion
          choco install cmake
```

rbsheth (commented):
I'm guessing you can get away with using only CMake here.

vthiery:
> @vthiery The impact on non-Hunter users is zero since I've disabled it by default. https://github.com/vthiery/cpp-statsd-client/pull/33/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR10
> 
> The argument is only needed since Hunter manages the C++ standard/other flags across all compile units.

It doesn't harm indeed but I'm skeptical about having to change a library (or its build config) to have a package manager working. Actually, shouldn't we avoid setting these 
```
set(CMAKE_CXX_STANDARD 11)
set(CMAKE_CXX_EXTENSIONS OFF)
```
and let the lib consumer set them?

rbsheth:
@vthiery Yes, that would be ideal. But setting those is one way to communicate to consumers that the library needs C++11 support. Another way is to set per-target requirements https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD. I would suggest setting these on test/other executables and refraining from setting on the library.

.github/workflows/win.yml
```diff
@@ -0,0 +1,18 @@
...
+  windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: dependencies
+        run: |
+          choco install cmake
+      - name: build
+        run: |
+          cmake -H . -B build -G "Visual Studio 15 2017" -A x64
```
rbsheth, line 14: Sorry, you'll need to remove spaces here
```suggestion
          cmake -H. -Bbuild -G "Visual Studio 15 2017" -A x64
```

.github/workflows/win.yml
```diff
@@ -0,0 +1,18 @@
...
+  windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: dependencies
+        run: |
+          choco install cmake
+      - name: build
+        run: |
+          cmake -H. -Bbuild -G "Visual Studio 15 2017" -A x64
```
rbsheth, line 14: Wow, actions doesn't like me either.
```suggestion
          cmake -S . -B build -G "Visual Studio 15 2017" -A x64
```

.github/workflows/win.yml
```diff
@@ -0,0 +1,18 @@
...
+  windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: dependencies
+        run: |
+          choco install cmake
+      - name: build
+        run: |
+          cmake -S . -B build -G "Visual Studio 15 2017" -A x64
```
rbsheth, line 14: Looks like only 2019 is installed. This should be it! https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#visual-studio-enterprise-2019

```suggestion
          cmake -S . -B build -G "Visual Studio 16 2019" -A x64
```

kevinkreiser:
getting close now, should i remove the hunter change for the time being so as not to couple it to windows support? then we can pr it separately?

kevinkreiser:
haha i had to get a windows vm and install visual studio to actually get this working but it seems to pass now. ill clean it up and ping  back for review

CMakeLists.txt
```diff
@@ -5,9 +5,16 @@ project(cpp-statsd-client
         LANGUAGES CXX
         DESCRIPTION "A header-only StatsD client implemented in C++"
         HOMEPAGE_URL "https://github.com/vthiery/cpp-statsd-client")
+
+option(CPP_STATSD_STANDALONE "Allows configuration of targets for verifying library functionality" ON)
+if(${CPP_STATSD_STANDALONE})
```
kevinkreiser, line 10: ive added this option to disable the creation of any special  targets for testing the library in standalone mode. this is useful when a project vendors this library in their cmake as they can simply do:

```cmake
SET(BUILD_TESTS OFF CACHE BOOL "ignore test targets" FORCE)
add_subdirectory(${CMAKE_SOURCE_DIR}/thirdparty/cpp-statsd-client ${CMAKE_BINARY_DIR}/thirdparty/cpp-statsd-client)
```

and then they can just depend on the `cpp-statsd-client` target in their targets

CMakeLists.txt
```diff
@@ -54,13 +64,19 @@ install(FILES "${PROJECT_BINARY_DIR}/${PROJECT_NAME}Config.cmake"
...
-# The test targets
-add_executable(testStatsdClient ${CMAKE_CURRENT_SOURCE_DIR}/tests/testStatsdClient.cpp)
-target_compile_options(testStatsdClient PRIVATE -Wall -Wextra -pedantic -Werror)
-target_include_directories(testStatsdClient PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/tests)
-target_link_libraries(testStatsdClient ${PROJECT_NAME})
+if(ENABLE_TESTS)
+  # The test targets
+  add_executable(testStatsdClient ${CMAKE_CURRENT_SOURCE_DIR}/tests/testStatsdClient.cpp)
+  if(WIN32)
+    target_compile_options(testStatsdClient PRIVATE -W4 -WX /external:W0)
```
kevinkreiser, line 71: we have to use different warning levels for windows. their pedantic warning level is too pedantic :smile:

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -67,6 +67,7 @@ class StatsdClient {
 
     StatsdClient(const StatsdClient&) = delete;
     StatsdClient& operator=(const StatsdClient&) = delete;
+    StatsdClient(StatsdClient&&) = default;
```
kevinkreiser, line 70: fixes a warning in windows builds

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -1,20 +1,39 @@
...
+
+#ifdef _WIN32
+using SOCKET_TYPE = SOCKET;
+constexpr SOCKET_TYPE k_invalidSocket{INVALID_SOCKET};
+#define SOCKET_ERRNO WSAGetLastError()
+#else
+using SOCKET_TYPE = int;
+constexpr SOCKET_TYPE k_invalidSocket{-1};
+#define SOCKET_ERRNO errno
+#endif
```
kevinkreiser, line 35: i found these helpful so i didnt have to do a bunch of `ifdef`ing inline in the code below

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -118,10 +141,35 @@ class UDPSender final {
 
     //! Error message (optional string)
     std::string m_errorMessage;
+};
+
+#ifdef _WIN32
```
kevinkreiser, line 146: in windows you have to initialize and tear down `WSA` before doing socket operations. so i wrapped it in a threadsafe singleton idiom and used the RAII idiom to take care of the teardown

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -118,10 +141,35 @@ class UDPSender final {
...
+        m_ok = WSAStartup(MAKEWORD(2, 2), &wsa) == 0;
+    }
+    bool m_ok;
 };
+}  // namespace detail
+#endif
+
+inline bool isValidSocket(const SOCKET_TYPE socket) {
+    return socket != k_invalidSocket;
+}
```
kevinkreiser, line 172: maybe this should also be in the detail namespace.. ill move that

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -165,12 +213,16 @@ inline UDPSender::~UDPSender() {
 
     // If we're running a background thread tell it to stop
     if (m_batchingThread.joinable()) {
-        m_mustExit.store(true, std::memory_order_acq_rel);
+        m_mustExit.store(true, std::memory_order_release);
```
kevinkreiser, line 216: i had to change these acquire on load and release on store otherwise windows would give back "invalid memory order"

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -165,12 +213,16 @@ inline UDPSender::~UDPSender() {
...
         m_batchingThread.join();
     }
 
     // Cleanup the socket
+#ifdef _WIN32
+    closesocket(m_socket);
+#else
     close(m_socket);
+#endif
```
kevinkreiser, line 225: since we do this alot i think ill move this up into a macro as well

tests/StatsdServer.hpp
```diff
@@ -1,19 +1,28 @@
 #ifndef STATSD_SERVER_HPP
 #define STATSD_SERVER_HPP
 
-#include <netinet/in.h>
-#include <sys/socket.h>
-#include <unistd.h>
+// It might make sense to include this test class in the UDPSender header
+// it includes most of the cross platform defines etc that we need for socket io
+#include "cpp-statsd-client/UDPSender.hpp"
```
kevinkreiser, line 6: because this class shares so much in common with UDPSender i didnt want to duplicate that code so i just included the header here. honestly i think it would kind of make sense to throw this class into the detail namespace in the UDPSender header but no worries if others feel strongly that we shouldnt.

tests/testStatsdClient.cpp
```diff
@@ -31,10 +31,19 @@ void mock(StatsdServer& server, std::vector<std::string>& messages) {
     } while (server.errorMessage().empty() && !messages.back().empty());
 }
 
-void throwOnError(const StatsdClient& client, bool expectEmpty = true, const std::string& extraMessage = "") {
-    if (client.errorMessage().empty() != expectEmpty) {
-        std::cerr << (expectEmpty ? client.errorMessage() : extraMessage) << std::endl;
-        throw std::runtime_error(expectEmpty ? client.errorMessage() : extraMessage);
+template <typename SocketWrapper>
```
kevinkreiser, line 34: i templated this so i could use it with the client or the server as they both adhere to the same spec in terms of getting error strings

tests/testStatsdClient.cpp
```diff
@@ -162,16 +164,21 @@ void testSendRecv(uint64_t batchSize, uint64_t sendInterval) {
...
+        testReconfigure();
+        // no batching
+        testSendRecv(0, 0);
+        // background batching
+        testSendRecv(32, 1000);
+        // manual flushing of batches
+        testSendRecv(16, 0);
+    } catch (const std::exception& e) {
+        std::cerr << e.what() << std::endl;
+        return EXIT_FAILURE;
```
kevinkreiser, line 180: this happens automatically anyway with an uncaught exception but i think it wasnt printing the `what` automatically. i can revert this though as im not 100% sure its needed

kevinkreiser:
@vthiery this is ready for review! windows CI is up and running (though much slower than linux) and the tests are passing on both platforms with maximal (but still sane) warnings turned on.

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -1,20 +1,41 @@
...
+using SOCKET_TYPE = SOCKET;
+constexpr SOCKET_TYPE k_invalidSocket{INVALID_SOCKET};
+#define SOCKET_ERRNO WSAGetLastError()
+#define SOCKET_CLOSE closesocket
+#else
+using SOCKET_TYPE = int;
+constexpr SOCKET_TYPE k_invalidSocket{-1};
+#define SOCKET_ERRNO errno
+#define SOCKET_CLOSE close
+#endif
```
kevinkreiser, line 37: i found these helpful so i didnt have to do a bunch of ifdefing inline in the code below

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -118,10 +143,37 @@ class UDPSender final {
...
+};
+
+namespace detail {
+
+inline bool isValidSocket(const SOCKET_TYPE socket) {
+    return socket != k_invalidSocket;
+}
+
+#ifdef _WIN32
+struct WinSockSingleton {
```
kevinkreiser, line 155: in windows you have to initialize and tear down WSA before doing socket operations. so i wrapped it in a threadsafe singleton idiom and used the RAII idiom to take care of the teardown

CMakeLists.txt
```diff
@@ -5,9 +5,19 @@ project(cpp-statsd-client
...
         HOMEPAGE_URL "https://github.com/vthiery/cpp-statsd-client")
+
+option(CPP_STATSD_STANDALONE "Allows configuration of targets for verifying library functionality" ON)
+option(ENABLE_TESTS "Build tests" ON)
+option(ENABLE_COVERAGE "Build with coverage instrumentalisation" OFF)
+
+if(NOT CPP_STATSD_STANDALONE)
+  set(ENABLE_TESTS OFF)
+  set(ENABLE_COVERAGE OFF)
+endif()
```
kevinkreiser, line 16: ive added this option to disable the creation of any special  targets for testing the library in standalone mode. this is useful when a project vendors this library in their cmake as they can simply do:

```cmake
SET(CPP_STATSD_STANDALONE OFF CACHE BOOL "ignore test targets" FORCE)
add_subdirectory(${CMAKE_SOURCE_DIR}/thirdparty/cpp-statsd-client ${CMAKE_BINARY_DIR}/thirdparty/cpp-statsd-client)
```

and then they can just depend on the `cpp-statsd-client` target in their targets

kevinkreiser: with out this its highly likely that the targets we create will collide with downstream targets since they are named `check` and `coverage` which are pretty generic

kevinkreiser:
@vthiery this is ready for review, let me know if i can make that easier in any way, i know its a lot

vthiery:
> @vthiery this is ready for review, let me know if i can make that easier in any way, i know its a lot

Thanks for the work and the many comments @kevinkreiser ! I'll try to review this one asap

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -251,16 +309,24 @@ inline bool UDPSender::initialize() noexcept {
...
 inline void UDPSender::sendToDaemon(const std::string& message) noexcept {
     // Try sending the message
-    const long int ret{
-        sendto(m_socket, message.data(), message.size(), 0, (struct sockaddr*)&m_server, sizeof(m_server))};
+    const auto ret = sendto(m_socket,
+                            message.data(),
+#ifdef _WIN32
+                            static_cast<int>(message.size()),
+#else
+                            message.size(),
```
kevinkreiser, line 317: windows takes a signed int while linux takes size_t

include/cpp-statsd-client/StatsdClient.hpp
vthiery: Why `default` and not `delete`?

include/cpp-statsd-client/UDPSender.hpp
vthiery: Hmm ok. Do you have more info about why this segfault occurs?

tests/StatsdServer.hpp
vthiery: > it would kind of make sense to throw this class into the detail namespace in the UDPSender header

Not sure I understand what should be included where.

But adding the `UDPSender` header here is ok to avoid duplicating code.

include/cpp-statsd-client/UDPSender.hpp
kevinkreiser: seems to occur when the `load` gets to the `release` part. here are the messages i see in visual studio in order:

image

image

image

kevinkreiser: so it hits an assertion in debug mode which would give me the intuition that we are doing something specifically wrong. then i click ignore to see what will happen next and it throws an exception so i look at the call stack and we can see it was in the lambda defined in the constructor. anyway when i first saw this i went and read the docs and found this bit:

https://en.cppreference.com/w/cpp/atomic/memory_order

> **Release-Acquire ordering**
> If an atomic store in thread A is tagged memory_order_release and an atomic load in thread B from the same variable is tagged memory_order_acquire, all memory writes (non-atomic and relaxed atomic) that happened-before the atomic store from the point of view of thread A, become visible side-effects in thread B. That is, once the atomic load is completed, thread B is guaranteed to see everything thread A wrote to memory.

This sounded just like what we were looking for so i changed the implementation to match that suggestion.

tests/StatsdServer.hpp
kevinkreiser: ah sorry for the confusion comment, i was again suggesting that we could just take the tiny bit of code that exists in this header and put it inside the `detail` namespace inside of the `UDPSender.hpp` header and completely delete this header. having a `UDP.hpp` with both a sender and receiver makes organizational sense to me. but again if you prefer to keep this test code out of the distributed header no problem. it was just a suggestion

include/cpp-statsd-client/StatsdClient.hpp
kevinkreiser: turns out that was only a warning when i turned on `/Wall` which was way too far of a level to turn on anyway so i can just remove it

include/cpp-statsd-client/UDPSender.hpp
kevinkreiser: ah here is the answer in the docs for load and store:

https://en.cppreference.com/w/cpp/atomic/atomic/load
> order must be one of std::memory_order_relaxed, std::memory_order_consume, std::memory_order_acquire or std::memory_order_seq_cst. Otherwise the behavior is undefined.

https://en.cppreference.com/w/cpp/atomic/atomic/store
> order must be one of std::memory_order_relaxed, std::memory_order_release or std::memory_order_seq_cst. Otherwise the behavior is undefined.

vthiery:
> haha i had to get a windows vm and install visual studio to actually get this working

What's the most straightforward setup to test this (when you run on linux)? we could add this to the README

kevinkreiser:
@vthiery im not exactly sure what you mean by: "when you run on linux". do you mean how to check windows builds work on linux? for me i created a virtual box vm of windows and downloaded visual studio and basically ran it on windows natively that way. it will be the same for any windows developer. you simply open the project directory as a cmake project in visual studio and run the test target and it just works. do you want me to document that in the readme?

the comment that you quoted was me laughing at myself for thinking i could debug the windows build via github actions CI logs. in the end it was just too slow to do it that way. so instead of taking forever i just installed a vm and worked in windows directly.

does that make sense?

vthiery:
> i created a virtual box vm of windows and downloaded visual studio and basically ran it on windows natively 

:+1:

vthiery (approved):

PR #34, opened 2021-06-25
Hunter and Windows support

rbsheth:
Add support for the Hunter package manager. Also add Windows support.


vthiery:
Since Windows support has been implemented in #33, we're left with Hunter support. If what I suggested in https://github.com/vthiery/cpp-statsd-client/pull/33#issuecomment-873090098 is feasible, I'd go with that. Otherwise, I'm fine adding a flag to support Hunter.

rbsheth:
@vthiery Great! I've updated this PR to just set those properties on the test target, so no need for a Hunter-specific flag.

PR #35, opened 2021-06-29
breakfix for ODR

kevinkreiser:
if you include this header in multiple compilation units you will violate the ODR and your compiler will tell you that its multiply defined. we inline it to fix this like the rest of the library

### DIFF
--- a/include/cpp-statsd-client/StatsdClient.hpp
+++ b/include/cpp-statsd-client/StatsdClient.hpp
@@ -153,7 +153,7 @@ private:
 };
 
 namespace detail {
-std::string sanitizePrefix(std::string prefix) {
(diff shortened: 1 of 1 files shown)


include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -153,7 +153,7 @@ class StatsdClient {
 };
 
 namespace detail {
-std::string sanitizePrefix(std::string prefix) {
+inline std::string sanitizePrefix(std::string prefix) {
```
kevinkreiser, line 156: sorry for this oversight earlier :facepalm:

vthiery (approved):
Thanks for the quick fix!

PR #38, opened 2021-07-21
update linux and windows badges

kevinkreiser:
i tried updating the badges in the readme to match the actual workflows. i think its nice to advertize windows support right at the top of the readme


vthiery (approved):
Thanks!

PR #39, opened 2021-08-16
Allow float gauges

messiaen:
@vthiery Currently only int values can be given for gauges due to the functions definitions.  I'd like to be able to send floats.

### DIFF
--- a/include/cpp-statsd-client/StatsdClient.hpp
+++ b/include/cpp-statsd-client/StatsdClient.hpp
@@ -4,8 +4,10 @@
 #include <cpp-statsd-client/UDPSender.hpp>
 #include <cstdint>
 #include <cstdio>
+#include <iomanip>
 #include <memory>
 #include <random>
+#include <sstream>
 #include <string>
 #include <vector>
 
@@ -63,7 +65,8 @@ public:
                  const uint16_t port,
                  const std::string& prefix,
                  const uint64_t batchsize = 0,
-                 const uint64_t sendInterval = 1000) noexcept;
+                 const uint64_t sendInterval = 1000,
+                 const unsigned int gaugePrecision = 4) noexcept;
 
     StatsdClient(const StatsdClient&) = delete;
     StatsdClient& operator=(const StatsdClient&) = delete;
@@ -78,7 +81,8 @@ public:
                    const uint16_t port,
                    const std::string& prefix,
                    const uint64_t batchsize = 0,
-                   const uint64_t sendInterval = 1000) noexcept;
+                   const uint64_t sendInterval = 1000,
+                   const unsigned int gaugePrecision = 4) noexcept;
 
     //! Returns the error message as an std::string
     const std::string& errorMessage() const noexcept;
@@ -100,8 +104,9 @@ public:
                const std::vector<std::string>& tags = {}) const noexcept;
 
     //! Records a gauge for the key, with a given value, at a given frequency rate
+    template <typename T>
     void gauge(const std::string& key,
-               const unsigned int value,
+               const T value,
                float frequency = 1.0f,
                const std::vector<std::string>& tags = {}) const noexcept;
 
@@ -130,8 +135,9 @@ private:
     // @{
 
     //! Send a value for a key, according to its type, at a given frequency
+    template <typename T>
     void send(const std::string& key,
-              const int value,
+              const T value,
               const char* type,
               float frequency,
               const std::vector<std::string>& tags) const noexcept;
@@ -150,6 +156,9 @@ private:
 
     //! The buffer string format our stats before sending them
     mutable std::string m_buffer;
+
+    //! Fixed floating point precision of gauges
+    unsigned int m_gaugePrecision;
 };
 
 namespace detail {
@@ -172,8 +181,11 @@ inline StatsdClient::StatsdClient(const std::string& host,
                                   const uint16_t port,
                                   const std::string& prefix,
                                   const uint64_t batchsize,
-                                  const uint64_t sendInterval) noexcept
-    : m_prefix(detail::sanitizePrefix(prefix)), m_sender(new UDPSender{host, port, batchsize, sendInterval}) {
+                                  const uint64_t sendInterval,
+                                  const unsigned int gaugePrecision) noexcept
+    : m_prefix(detail::sanitizePrefix(prefix)),
+      m_sender(new UDPSender{host, port, batchsize, sendInterval}),
+      m_gaugePrecision(gaugePrecision) {
     // Initialize the random generator to be used for sampling
     seed();
     // Avoid re-allocations by reserving a generous buffer
@@ -184,9 +196,11 @@ inline void StatsdClient::setConfig(const std::string& host,
                                     const uint16_t port,
                                     const std::string& prefix,
                                     const uint64_t batchsize,
-                                    const uint64_t sendInterval) noexcept {
+                                    const uint64_t sendInterval,
+                                    const unsigned int gaugePrecision) noexcept {
     m_prefix = detail::sanitizePrefix(prefix);
     m_sender.reset(new UDPSender(host, port, batchsize, sendInterval));
+    m_gaugePrecision = gaugePrecision;
 }
 
 inline const std::string& StatsdClient::errorMessage() const noexcept {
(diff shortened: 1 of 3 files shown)


include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -209,35 +242,73 @@ inline void StatsdClient::count(const std::string& key,
...
+                               const std::vector<std::string> &tags) const noexcept {
+    send(key, std::to_string(value), type, frequency, tags);
+}
+
+inline void StatsdClient::send(const std::string &key,
+                               const float value,
+                               const char *type,
+                               float frequency,
+                               const std::vector<std::string> &tags) const noexcept {
+    send(key, std::to_string(value), type, frequency, tags);
```
kevinkreiser, line 304: @messiaen rather than add `to_string` everywhere could we just `template` the send function so it could take either float or int?

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -136,6 +148,27 @@ class StatsdClient {
...
+              const char *type,
+              float frequency,
+              const std::vector<std::string> &tags) const noexcept;
+
+    //! Send a value for a key, according to its type, at a given frequency
+    void send(const std::string &key,
+              const std::string &value,
+              const char *type,
+              float frequency,
+              const std::vector<std::string> &tags) const noexcept;
```
kevinkreiser, line 170: same point as below, why not use templates?

tests/testStatsdClient.cpp
```diff
@@ -110,10 +110,15 @@ void testSendRecv(uint64_t batchSize, uint64_t sendInterval) {
         throwOnError(client);
 
         // Record a gauge "titi" to 3
-        client.gauge("titi", 3);
+        client.gauge("titi", 3u);
         throwOnError(client);
         expected.emplace_back("sendRecv.titi:3|g");
 
+        // Record a gauge "titifloat" to 123.456
+        client.gauge("titifloat", -123.456f);
```
kevinkreiser, line 118: my biggest question about this PR is about precision. single precision floats are notorious for storing essentially garbage noise after the 6th (or so) digit. converting and sending these bytes seems wasteful. maybe we should be using a fixed precision for floating point numbers? or maybe the extra noise doesnt bother anyone else?

messiaen: Added a precision config value in StatusdClient.

include/cpp-statsd-client/StatsdClient.hpp
messiaen: Using template now.

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -168,25 +176,30 @@ constexpr char METRIC_TYPE_TIMING[] = "ms";
...
-    : m_prefix(detail::sanitizePrefix(prefix)), m_sender(new UDPSender{host, port, batchsize, sendInterval}) {
+                                  const uint64_t sendInterval,
+                                  const unsigned int gaugePrecision) noexcept
+    : m_prefix(detail::sanitizePrefix(prefix)),
+      m_sender(new UDPSender{ host, port, batchsize, sendInterval }) {
     // Initialize the random generator to be used for sampling
     seed();
     // Avoid re-allocations by reserving a generous buffer
     m_buffer.reserve(256);
+    this->gaugePrecision = gaugePrecision;
```
kevinkreiser, line 191: a couple notes on style:

it seems that your formatter is doing some strange things with the white space which is makign the diff hard to read. we are using clangformat here. also notice that we use `hungarian` notation, so that variable should be `m_guagePrecision` since its a member variable. also, i would suggest not using `this->` its superfluous (and looks like java :trollface: )

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -209,35 +222,38 @@ inline void StatsdClient::count(const std::string& key,
                                 const int delta,
                                 float frequency,
                                 const std::vector<std::string>& tags) const noexcept {
-    return send(key, delta, detail::METRIC_TYPE_COUNT, frequency, tags);
+    send(key, delta, detail::METRIC_TYPE_COUNT, frequency, tags);
```
kevinkreiser, line 225: ha good catch!

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -130,18 +134,22 @@ class StatsdClient {
...
-              const char* type,
+    template <typename T>
+    void send(const std::string &key,
+              const T value,
+              const char *type,
               float frequency,
-              const std::vector<std::string>& tags) const noexcept;
-
+              const std::vector<std::string> &tags,
+              unsigned int precision = 0) const noexcept;
```
kevinkreiser, line 143: precision is a member variable so you dont need to change this method signature, the method has already access to the member variable

include/cpp-statsd-client/StatsdClient.hpp
messiaen: - Ran clang-format again.
- Changed instance variable name.

kevinkreiser (approved):
the code looks good to me but i defer to @vthiery. also we need to allow CI to run too just to make sure

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -78,20 +81,19 @@ class StatsdClient {
...
 
     //! Returns the error message as an std::string
     const std::string& errorMessage() const noexcept;
 
     //! Increments the key, at a given frequency rate
-    void increment(const std::string& key,
-                   float frequency = 1.0f,
-                   const std::vector<std::string>& tags = {}) const noexcept;
+    void increment(const std::string& key, float frequency = 1.0f, const std::vector<std::string>& tags = {}) const
+        noexcept;
```
kevinkreiser, line 92: nothing changed here but there is still a diff, i guess you must have a different clangformat version or a different default line width? any chance you can clean up these whitespace changes so that the diff is just focused on actual changes?

include/cpp-statsd-client/StatsdClient.hpp
vthiery: Indeed :)
The same can be done to `decrement` and `increment`.

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -172,69 +179,73 @@ inline StatsdClient::StatsdClient(const std::string& host,
...
                                   const uint64_t batchsize,
-                                  const uint64_t sendInterval) noexcept
+                                  const uint64_t sendInterval,
+                                  const unsigned int gaugePrecision) noexcept
     : m_prefix(detail::sanitizePrefix(prefix)), m_sender(new UDPSender{host, port, batchsize, sendInterval}) {
     // Initialize the random generator to be used for sampling
     seed();
     // Avoid re-allocations by reserving a generous buffer
     m_buffer.reserve(256);
+    m_gaugePrecision = gaugePrecision;
```
vthiery, line 189: You can do like `m_prefix` and `m_sender` line 184:
```cpp
: m_prefix(detail::sanitizePrefix(prefix)), m_sender(new UDPSender{host, port, batchsize, sendInterval}), m_gaugePrecision(gaugePrecision)
```

tests/testStatsdClient.cpp
```diff
@@ -114,6 +114,11 @@ void testSendRecv(uint64_t batchSize, uint64_t sendInterval) {
         throwOnError(client);
         expected.emplace_back("sendRecv.titi:3|g");
 
+        // Record a gauge "titifloat" to 123.456
+        client.gauge("titifloat", 123.456);
+        throwOnError(client);
+        expected.emplace_back("sendRecv.titifloat:123.456|g");
```
vthiery, line 120: We should check that the precision is satisfied:
```suggestion
        // Record a gauge "titifloat" to 123.456789 with precision 3
        client.gauge("titifloat", 123.456789);
        throwOnError(client);
        expected.emplace_back("sendRecv.titifloat:123.457|g");
```

vthiery (changes requested):
Thanks for the PR! Let's just tweak the test a bit (or add one, as you wish), fix the style and then it should be ok :+1:

vthiery:
> the code looks good to me but i defer to @vthiery. also we need to allow CI to run too just to make sure

Thanks for the review @kevinkreiser ! I allowed the CI to run :+1:

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -63,7 +65,8 @@ class StatsdClient {
                  const uint16_t port,
                  const std::string& prefix,
                  const uint64_t batchsize = 0,
-                 const uint64_t sendInterval = 1000) noexcept;
+                 const uint64_t sendInterval = 1000,
+                 const unsigned int gaugePrecision = 4) noexcept;
```
vthiery, line 69: We could also add a quick note about the setting float precision (a line or two is enough) in the README, like we did for the other parameters.

messiaen:
Thanks @vthiery and @kevinkreiser! 
- Fixed formatting
- Added note about gaugePrecision in code examples in README
- tweaked test.

tests/testStatsdClient.cpp
messiaen: Made this change manually.

vthiery (approved):
LGTM
@kevinkreiser - want to have another look before I merge it?

dmitrifried:
Hi @vthiery, are you planning on making a new release version with this commit? Right now our cmake dependency will have to point to the commit hash to get the float guages.

vthiery:
> Hi @vthiery, are you planning on making a new release version with this commit? Right now our cmake dependency will have to point to the commit hash to get the float guages.

Hi @dmitrifried 
I'll release a v2 today or tomorrow. I just need to prepare a quick migration guide since we've been making quite a lot of changes lately.

vthiery:
> > Hi @vthiery, are you planning on making a new release version with this commit? Right now our cmake dependency will have to point to the commit hash to get the float guages.
> 
> Hi @dmitrifried
> I'll release a v2 today or tomorrow. I just need to prepare a quick migration guide since we've been making quite a lot of changes lately.

@dmitrifried I just release [v2.0.0](https://github.com/vthiery/cpp-statsd-client/releases/tag/v2.0.0)

dmitrifried:
@vtheiry Thanks for the quick work!

PR #43, opened 2022-05-15
Add `-Wsign-conversion -Wsign-promo` compiler flags

dsze:
I'm trying to use cpp-statsd-client in a different project that uses stricter compiler warning flags on Linux.  The changes in this PR were required to fix the issues found by the flags.


kevinkreiser:
These flags are quite strict but your changes look correct to me and wider compatibility of the project with more downstream projects is a good thing!

Looks like the windows build is broken through no fault of your own. We will have to see if they changed the version available in CI

kevinkreiser (approved):
Looks good but I'd wait for a review from @vthiery first before merging

kevinkreiser:
@dsze many thanks for updating to the later version of vs for the windows builds looks like everything is passing!

PR #44, opened 2022-06-07
feat: support custom metric types

dsze:
Other statsd backends (e.g. Datadog, netdata) support metric types beyond
those supported by the original Etsy statsd daemon.  For example:

* Histograms (type "h") which are like Timers, but with different units.

* Dictionaries (type "d") which are like Sets but also record a count for
  each unique value.

Since these are non-standard, add a `custom()` method, where the metric
type is passed in by the caller, assuming that they know the capabilities
of the backend they are using.


dsze:
See https://github.com/vthiery/cpp-statsd-client/pull/45 instead.  Had to close this PR because it was created off of `master` on my fork, making it difficult to update.

PR #45, opened 2022-06-08
feat: support custom metric types

dsze:
Add a `custom()` method where the metric type is passed in by the caller,
to allow using statsd backends that support metric types beyond those
supported by the original Etsy statsd daemon.


kevinkreiser (approved):
thank you for updating the docs with examples!

PR #47, opened 2023-01-24
Fix FreeBSD compile error

jeremiahpslewis:
Found while building Valhalla for Julia: https://github.com/JuliaPackaging/Yggdrasil/pull/6127

### DIFF
--- a/include/cpp-statsd-client/UDPSender.hpp
+++ b/include/cpp-statsd-client/UDPSender.hpp
@@ -14,7 +14,7 @@
 #endif
 
 #if defined(__FreeBSD__)
-    #include <netinet/in.h>
+#include <netinet/in.h>
 #endif    
     
 #include <atomic>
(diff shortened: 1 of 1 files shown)


kevinkreiser (approved):
it's too bad we can have infinite  CI then we could have a build for all platforms. we can just take your word for this one though I think

kevinkreiser:
@vthiery what do you think? I think the ifdef makes it harmless since all but freebsd will make use of this code. IMHO it's not a strong enough case to build a CI job for it.

vthiery:
> @vthiery what do you think? I think the ifdef makes it harmless since all but freebsd will make use of this code. IMHO it's not a strong enough case to build a CI job for it.

I agree that setting up a CI for this would probably be overkill at this point.

vthiery (approved):

PR #50, opened 2024-04-02
Fix include install directory

paulbovbel:
Current implementation installs into ${INSTALL_DIR}/include/include/cpp-stats-d


kevinkreiser (approved):
ha! i just had another project i was working on today where i had and fixed this same mistake! thanks!

PR #52, opened 2025-03-09
fix: include incoming message size when checking against batch size

kwvg:
## Expected behavior

* Create `StatsdClient` with `batchsize`
* Push a few messages that are under `batchsize`, remains in same batch
* Push a message that would go over `batchsize` if appended,  gets put into a **new** batch
* Trigger flush, two batches are created, both **under** `batchsize`

## Actual behavior

* Create `StatsdClient` with `batchsize`
* Push a few messages that are under `batchsize`, remains in same batch
* Push a message that would go over `batchsize` if appended,  gets put into **same** batch
* Flush triggered, one batch is created **exceeding** `batchsize`

## Steps to reproduce

* Revert e3bffa355c069b89968877adc8d3933cf27d83d9 and run tests, a hyphen (`-`) is used to indicate a batch

### DIFF
--- a/include/cpp-statsd-client/UDPSender.hpp
+++ b/include/cpp-statsd-client/UDPSender.hpp
@@ -246,8 +246,8 @@ inline void UDPSender::queueMessage(const std::string& message) noexcept {
     // We aquire a lock but only if we actually need to (i.e. there is a thread also accessing the queue)
     auto batchingLock =
         m_batchingThread.joinable() ? std::unique_lock<std::mutex>(m_batchingMutex) : std::unique_lock<std::mutex>();
-    // Either we don't have a place to batch our message or we exceeded the batch size, so make a new batch
-    if (m_batchingMessageQueue.empty() || m_batchingMessageQueue.back().length() > m_batchsize) {
+    // Either we don't have a place to batch our message or we are about to exceed the batch size, so make a new batch
+    if (m_batchingMessageQueue.empty() || m_batchingMessageQueue.back().size() + message.size() > m_batchsize) {
         m_batchingMessageQueue.emplace_back();
         m_batchingMessageQueue.back().reserve(m_batchsize + 256);
     }  // When there is already a batch open we need a separator when its not empty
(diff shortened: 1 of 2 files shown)


kevinkreiser:
@kwvg the documentation https://github.com/vthiery/cpp-statsd-client?tab=readme-ov-file#batching explicitly states that the batch size is not a hard limit (indeed it cant be without quite a bit of annoyance, eg. in the case where the batch size is so small a single message wouldnt fit). so i wouldnt characterize this change as a "fix" in this case. this is merely just more of an attempt to make batch size a hard limit right? but even still, the change here does not make it a hard limit specifically in the example i mentioned above. i also see that you keep the additional memory reservation when adding a new batch, which makes no sense if we are trying to stick to a hard limit.

at any rate, i think we should first discuss the merits of a hard limit vs the practical reality. perhaps we can throw or something if someone wants to make a batch size that is too small to be reasonable so as to avoid the pitfall i mentioned above? at present i dont feel strong about having a hard limit vs a soft limit. the reason why we went with a soft one though in the first place was simply because it was just quite a bit less difficult to adhere to logically :smile:

include/cpp-statsd-client/UDPSender.hpp
```diff
@@ -246,8 +246,8 @@ inline void UDPSender::queueMessage(const std::string& message) noexcept {
     // We aquire a lock but only if we actually need to (i.e. there is a thread also accessing the queue)
     auto batchingLock =
         m_batchingThread.joinable() ? std::unique_lock<std::mutex>(m_batchingMutex) : std::unique_lock<std::mutex>();
-    // Either we don't have a place to batch our message or we exceeded the batch size, so make a new batch
-    if (m_batchingMessageQueue.empty() || m_batchingMessageQueue.back().length() > m_batchsize) {
+    // Either we don't have a place to batch our message or we are about to exceed the batch size, so make a new batch
+    if (m_batchingMessageQueue.empty() || m_batchingMessageQueue.back().size() + message.size() > m_batchsize) {
         m_batchingMessageQueue.emplace_back();
         m_batchingMessageQueue.back().reserve(m_batchsize + 256);
```
kevinkreiser, line 252: if we really do want a hard limit we should add some fuzz to the batch size for overages:
```suggestion
        m_batchingMessageQueue.back().reserve(m_batchsize);
```

kevinkreiser:
linting step is broken we'll have to fix that on our end. you can ignore that for now :smile:

Issue #12, opened 2020-04-22
If the client is thread safe?


hh23485:
Hi, thanks for the rare cpp statsd client!

We'd like to use the client but my concern is about if the client is thread safe? 
if not do you have some suggestions to use the client in multi-threads web services?

vthiery:
Hi!

Unfortunately this client is not thread-safe. Nothing prevents a thread to alter the internal udp sender while another one is already sending metrics.

hh23485:
@vthiery hi, Thanks for your response!

About the un-thread safe client, do you have some suggestions to use it in web project?
Or we have to implements a local-free queue as a buffer to a singleton client consumer?

vthiery:
@hh23485 I suppose that, as long as you don't explicitly reconfigure or destroy the client while using it, it will be ok. In the context of a web project, one should first initialize/configure the client, the start spawning threads to treat requests concurrently, and make sure all the requests have been handled before destroying the client.

hh23485:
@vthiery Sure, initialize only once, and destroy the client at the end of the application.

But to be more clear, do you mean, it's safe to call the singleton client's `increment` or `timing` from multi threads? or I need to do my actions to only one thread call these functions at the same time?

vthiery:
> But to be more clear, do you mean, it's safe to call the singleton client's `increment` or `timing` from multi threads? or I need to do my actions to only one thread call these functions at the same time?

It should be. The only sensitive piece is batching (`m_batchingMessageQueue` ), but I defined a mutex (`m_batchingMutex`) and take locks whenever I manipulate the queue.

hh23485:
> > But to be more clear, do you mean, it's safe to call the singleton client's `increment` or `timing` from multi threads? or I need to do my actions to only one thread call these functions at the same time?
> 
> It should be. The only sensitive piece is batching (`m_batchingMessageQueue`), but I defined a mutex (`m_batchingMutex`) and take locks whenever I manipulate the queue.

That's great, thanks for your explain.
I'll test and give a feedback.

Thanks, again.

vthiery:
@hh23485 have you had any issue related thread-safety? otherwise, I'll close this issue.

hh23485:
@vthiery Sure, It looks good to me.

jakila:
If you came here because you wonder if this library is thread safe: Please have a look into this later PR: https://github.com/vthiery/cpp-statsd-client/pull/29
[...]

PR #53, opened 2025-04-17
Make send thread safe

AadityaRavindran:
Running sending metrics through TSan reveals data races when we try to clear the buffer. Adding a mutex lock makes it thread safe.

### DIFF
--- a/include/cpp-statsd-client/StatsdClient.hpp
+++ b/include/cpp-statsd-client/StatsdClient.hpp
@@ -286,6 +286,8 @@ inline void StatsdClient::send(const std::string& key,
     std::stringstream valueStream;
     valueStream << std::fixed << std::setprecision(m_gaugePrecision) << value;
 
+    // the thread keeps this buffer around and reuses it, clear should be O(1)
+    // and reserve should only have to do so the first time, after that, it's a no-op
     static thread_local std::string buffer;
     buffer.clear();
     buffer.reserve(256);
--- a/include/cpp-statsd-client/StatsdClient.hpp
+++ b/include/cpp-statsd-client/StatsdClient.hpp
@@ -6,7 +6,6 @@
 #include <cstdio>
 #include <iomanip>
 #include <memory>
-#include <mutex>
 #include <random>
 #include <sstream>
 #include <string>
@@ -163,12 +162,6 @@ private:
     //! The random number generator for handling sampling
     mutable std::mt19937 m_randomEngine;
 
-    //! The buffer string format our stats before sending them
-    mutable std::string m_buffer;
-
-    //! The mutex to lock m_buffer
-    mutable std::mutex m_buffer_mutex;
-
     //! Fixed floating point precision of gauges
     int m_gaugePrecision;
 };
@@ -200,8 +193,6 @@ inline StatsdClient::StatsdClient(const std::string& host,
       m_gaugePrecision(gaugePrecision) {
     // Initialize the random generator to be used for sampling
     seed();
-    // Avoid re-allocations by reserving a generous buffer
-    m_buffer.reserve(256);
 }
 
 inline void StatsdClient::setConfig(const std::string& host,
@@ -295,36 +286,36 @@ inline void StatsdClient::send(const std::string& key,
     std::stringstream valueStream;
     valueStream << std::fixed << std::setprecision(m_gaugePrecision) << value;
 
-    std::lock_guard<std::mutex> buffer_lock(m_buffer_mutex);
-    m_buffer.clear();
+    std::string buffer;
+    buffer.reserve(256);
 
-    m_buffer.append(m_prefix);
+    buffer.append(m_prefix);
     if (!m_prefix.empty() && !key.empty()) {
-        m_buffer.push_back('.');
+        buffer.push_back('.');
     }
 
-    m_buffer.append(key);
-    m_buffer.push_back(':');
-    m_buffer.append(valueStream.str());
-    m_buffer.push_back('|');
-    m_buffer.append(type);
+    buffer.append(key);
+    buffer.push_back(':');
+    buffer.append(valueStream.str());
+    buffer.push_back('|');
+    buffer.append(type);
 
     if (frequency < 1.f) {
-        m_buffer.append("|@0.");
-        m_buffer.append(std::to_string(static_cast<int>(frequency * 100)));
+        buffer.append("|@0.");
+        buffer.append(std::to_string(static_cast<int>(frequency * 100)));
     }
 
     if (!tags.empty()) {
-        m_buffer.append("|#");
+        buffer.append("|#");
         for (const auto& tag : tags) {
-            m_buffer.append(tag);
-            m_buffer.push_back(',');
+            buffer.append(tag);
+            buffer.push_back(',');
         }
-        m_buffer.pop_back();
+        buffer.pop_back();
     }
 
     // Send the message via the UDP sender
-    m_sender->send(m_buffer);
+    m_sender->send(buffer);
 }
 
 inline void StatsdClient::seed(unsigned int seed) noexcept {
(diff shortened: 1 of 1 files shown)


kevinkreiser:
i mentioned this over in an open issue about this point: https://github.com/vthiery/cpp-statsd-client/issues/12#issuecomment-2738576484

tldr; the client is fairly light weight, you can simply use a separate client for every thread

AadityaRavindran:
> i mentioned this over in an open issue about this point: [#12 (comment)](https://github.com/vthiery/cpp-statsd-client/issues/12#issuecomment-2738576484)
> 
> tldr; the client is fairly light weight, you can simply use a separate client for every thread

Just curious then why m_buffer was made mutable and memory is reserved in construction, instead of just a local buffer variable. https://github.com/vthiery/cpp-statsd-client/blob/6efdb168a23edeeef888e8146f32e0d2c4b657b5/include/cpp-statsd-client/StatsdClient.hpp#L199

Would a nice middle-ground be a way to turn on thread safety if someone needs it? Or the alternative could be to move m_buffer to a local variable in `StatsdClient::send` instead?

kevinkreiser:
everything was done with the idea of performance in mind. doing a reservation upfront and allowing us to modify it on send lets us reuse that chunk of memory and avoid de/reallocs. we want the client to be as fast as possible, since its basically like a logger but pushing logs over the network. no one wants a logger that they have to worry about perfwise everyone assumes its basically free. so yeah that, at least in my opinion, was a design goal

kevinkreiser:
yeah we totally could have opt-in thread safety. we could also just perf test this and see if modern compilers are smart enough to know when its use is in a single threaded context

AadityaRavindran:
Alright, I did some benchmarking. TLDR; Using local buffer is much better than std::mutex, and does not add much performance overhead over a mutable class variable, while providing thread safety.

This is the test script
```
#include <chrono>
#include <iostream>
#include <string>
#include <thread>
#include <vector>
#include "cpp-statsd-client/StatsdClient.hpp"

const std::string host = "127.0.0.1";
const int port = 8125;
Statsd::StatsdClient statsd(host, port, "test");

void sendMetrics(int thread_id, int iteration_count) {
    for (int i = 0; i < iteration_count; ++i) {
        int value = rand() % 100000;
        statsd.gauge("test-metric", value, 1.0);
        // std::cout << "Thread " << thread_id << " sent metric: test-metric = " << value << std::endl;
    }
}

int main(int argc, char* argv[]) {
    int num_threads = std::stoi(argv[1]);
    int iteration_count = std::stoi(argv[2]);
    std::srand(static_cast<unsigned int>(std::time(nullptr)));
    std::vector<std::thread> threads;
    auto start = std::chrono::high_resolution_clock::now();
    for (int i = 0; i < num_threads; ++i) {
        threads.emplace_back(sendMetrics, i, iteration_count);
    }
    for (auto& thread : threads) {
        thread.join();
    }
    auto end = std::chrono::high_resolution_clock::now();
    std::chrono::duration<double, std::milli> duration = end - start;
    std::cout << "Total metrics sent: " << iteration_count * num_threads << std::endl;
    std::cout << "Time taken: " << duration.count() << " ms" << std::endl;
    std::cout << "Metrics sent per second: " << (iteration_count * num_threads / (duration.count() / 1000.0)) << std::endl;
    return 0;
}
```

Here are the results:
Test 1: no mutex, 1 thread
```
user@group:~$ g++ -o statsd_example statsd_example.cpp -I./cpp-statsd-client/ -lpthread && ./statsd_example 1 1000000
Total metrics sent: 1000000
Time taken: 3673.82 ms
Metrics sent per second: 272196
```

Test 2: no mutex, 10 threads (causes data race issues leading to stat parsing issues - `Splitting ':', unable to parse metric: test.test-metric818100|g`)
```
user@group:~$ g++ -o statsd_example statsd_example.cpp -I./cpp-statsd-client/ -lpthread && ./statsd_example 10 100000
Total metrics sent: 1000000
Time taken: 863.359 ms
Metrics sent per second: 1.15827e+06
```

Test 3: std::mutex, 1 thread
```
user@group:~$ g++ -o statsd_example statsd_example.cpp -I./cpp-statsd-client/ -lpthread && ./statsd_example 1 1000000
Total metrics sent: 1000000
Time taken: 3778.6 ms
Metrics sent per second: 264649
```

Test 4: std::mutex, 10 threads (no data races, obviously)
```
user@group:~$ g++ -o statsd_example statsd_example.cpp -I./cpp-statsd-client/ -lpthread && ./statsd_example 10 100000
Total metrics sent: 1000000
Time taken: 6998.56 ms
Metrics sent per second: 142887
```

Test 5: Local string buffer, 1 thread
```
user@group:~$ g++ -o statsd_example statsd_example.cpp -I./cpp-statsd-client/ -lpthread && ./statsd_example 1 1000000
Total metrics sent: 1000000
Time taken: 4002.7 ms
Metrics sent per second: 249831
```

Test 6: Local string buffer, 10 threads (no data races)
```
user@group:~$ g++ -o statsd_example statsd_example.cpp -I./cpp-statsd-client/ -lpthread && ./statsd_example 10 100000
Total metrics sent: 1000000
Time taken: 830.395 ms
Metrics sent per second: 1.20425e+06

```

AadityaRavindran:
And if you ended up instantiating StatsdClient every time you sent it (with mutable std::string m_buffer) it would be 10x slower
```
user@group:~$ g++ -o statsd_example statsd_example.cpp -I./cpp-statsd-client/ -lpthread && ./statsd_example 1 1000000
Total metrics sent: 1000000
Time taken: 28197.7 ms
Metrics sent per second: 35463.9
user@group:~$ g++ -o statsd_example statsd_example.cpp -I./cpp-statsd-client/ -lpthread && ./statsd_example 10 100000
Total metrics sent: 1000000
Time taken: 4872.05 ms
Metrics sent per second: 205252
```

AadityaRavindran:
Note that all of this testing was done with a Telegraf server running.

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -291,35 +286,36 @@ inline void StatsdClient::send(const std::string& key,
     std::stringstream valueStream;
     valueStream << std::fixed << std::setprecision(m_gaugePrecision) << value;
 
-    m_buffer.clear();
+    std::string buffer;
+    buffer.reserve(256);
```
kevinkreiser, line 290: i still cant believe this would be faster than keeping the buffer around! allocating once in the constructor and then using clear (which is O(1)) should always be faster than this.

maybe we could just make a static thread local buffer to achieve the same result and avoid the allocations/deallocations? something like:

```suggestion
    // the thread keeps this buffer around and reuses it, clear should be O(1) and reserve should only have to do so the first time, after that its a no-op
    static thread_local std::string buffer;
    buffer.clear();
    buffer.reserve(256);
```

AadityaRavindran: It isn't though. Single threaded performance of allocating once in the constructor is 3673.82 ms, whereas local allocation is 4002.7 ms. And we can't trust the multi-threaded performance of allocating in the constructor because it does the wrong thing.

kevinkreiser: > And we can't trust the multi-threaded performance of allocating in the constructor because it does the wrong thing.

Yeah, I wasn't concerned about multithreaded perf because testing thread-unsafe code with threads makes no sense :smile:

AadityaRavindran: ... And this is the constructor allocation performance from today.
```
user@group:~$ g++ -o statsd_example statsd_example.cpp -I./cpp-statsd-client/ -lpthread && ./statsd_example 1 1000000
Total metrics sent: 1000000
Time taken: 2780.39 ms
Metrics sent per second: 359662
```

kevinkreiser:
@AadityaRavindran can you merge master into this branch

AadityaRavindran:
Thank you for the quick responses!

PR #54, opened 2025-04-21
try newer images for ci

kevinkreiser:
trying to get ci working again


kevinkreiser:
just fixing up CI, i dont think we need exhaustive review

PR #55, opened 2025-04-21
further threadsafety hardening

### LINKED ISSUE #12: If the client is thread safe?

kevinkreiser:
over in #12 we've clarified a few things to make the library threadsafe. there are a couple more little things to finish the job though largely the library is already threadsafe. this pr makes the use of the rng threadsafe by using thread local storage. additionally we remove the ability to change the internal state of the client via `setConfig` and `seed`. you're only option for configuration is up front or to make a new client which is certainly a clearer contract

the only thing left to do is fix the tests since the api has changed some of the tests are hard to keep in their current form

fixes #12

### DIFF
--- a/include/cpp-statsd-client/StatsdClient.hpp
+++ b/include/cpp-statsd-client/StatsdClient.hpp
@@ -270,7 +270,7 @@ inline void StatsdClient::send(const std::string& key,
     constexpr float epsilon{0.0001f};
     const bool isFrequencyOne = std::fabs(frequency - 1.0f) < epsilon;
     const bool isFrequencyZero = std::fabs(frequency) < epsilon;
-    if (isFrequencyZero || (!isFrequencyOne && (frequency < std::uniform_real_distribution<float>(0.f, 1.f)(rng())))) {
+    if (isFrequencyZero || (!isFrequencyOne && (frequency < std::uniform_real_distribution<float>(0.f, 1.f)(detail::rng())))) {
         return;
     }
 
--- a/include/cpp-statsd-client/StatsdClient.hpp
+++ b/include/cpp-statsd-client/StatsdClient.hpp
@@ -162,7 +162,7 @@ inline std::string sanitizePrefix(std::string prefix) {
     return prefix;
 }
 
-std::mt19937& rng(unsigned int seed = 0) {
+inline std::mt19937& rng(unsigned int seed = 0) {
     static thread_local std::mt19937 twister(seed);
     return twister;
 }
@@ -186,17 +186,6 @@ inline StatsdClient::StatsdClient(const std::string& host,
       m_gaugePrecision(gaugePrecision) {
     // Initialize the random generator to be used for sampling
     detail::rng(seed);
-}
-
-inline void StatsdClient::setConfig(const std::string& host,
-                                    const uint16_t port,
-                                    const std::string& prefix,
-                                    const uint64_t batchsize,
-                                    const uint64_t sendInterval,
-                                    const int gaugePrecision) noexcept {
-    m_prefix = detail::sanitizePrefix(prefix);
-    m_sender.reset(new UDPSender(host, port, batchsize, sendInterval));
-    m_gaugePrecision = gaugePrecision;
 }
 
 inline const std::string& StatsdClient::errorMessage() const noexcept {
--- a/include/cpp-statsd-client/StatsdClient.hpp
+++ b/include/cpp-statsd-client/StatsdClient.hpp
@@ -4,14 +4,38 @@
 #include <cpp-statsd-client/UDPSender.hpp>
 #include <cstdint>
 #include <cstdio>
+#include <functional>
 #include <iomanip>
 #include <memory>
 #include <random>
 #include <sstream>
 #include <string>
+#include <utility>
 #include <vector>
 
 namespace Statsd {
+
+namespace detail {
+inline std::string sanitizePrefix(std::string prefix) {
+    // For convenience we provide the dot when generating the stat message
+    if (!prefix.empty() && prefix.back() == '.') {
+        prefix.pop_back();
+    }
+    return prefix;
+}
+
+inline float rand(unsigned int seed) {
+    static thread_local std::mt19937 twister(seed);
+    static thread_local std::uniform_real_distribution<float> dist(0.f, 1.f);
+    return dist(twister);
+}
+
+// All supported metric types
+constexpr char METRIC_TYPE_COUNT[] = "c";
+constexpr char METRIC_TYPE_GAUGE[] = "g";
+constexpr char METRIC_TYPE_TIMING[] = "ms";
+constexpr char METRIC_TYPE_SET[] = "s";
+}  // namespace detail
 
 /*!
  *
@@ -27,8 +51,8 @@ namespace Statsd {
  * nor prepend one to the key
  *
  * The sampling frequency is specified per call and uses a
- * random number generator to determine whether or not the stat
- * will be recorded this time or not.
+ * random number generator (optionally user-specified) to
+ * determine whether the stat will be recorded this time or not.
  *
  * The top level configuration includes 2 optional parameters
  * that determine how the stats are delivered to statsd. These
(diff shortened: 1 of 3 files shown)


include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -4,15 +4,39 @@
...
+    static thread_local std::uniform_real_distribution<float> dist(0.f, 1.f);
+    return dist(twister);
+}
+
+// All supported metric types
+constexpr char METRIC_TYPE_COUNT[] = "c";
+constexpr char METRIC_TYPE_GAUGE[] = "g";
+constexpr char METRIC_TYPE_TIMING[] = "ms";
+constexpr char METRIC_TYPE_SET[] = "s";
+}  // namespace detail
```
kevinkreiser, line 38: moved this up above the constructor for the class so i could make use of the random function as a default argument

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -66,7 +95,9 @@ class StatsdClient {
                  const std::string& prefix,
                  const uint64_t batchsize = 0,
                  const uint64_t sendInterval = 1000,
-                 const int gaugePrecision = 4) noexcept;
+                 const int gaugePrecision = 4,
+                 FrequencyFunc frequencyFunc =
+                     std::bind(detail::rand, std::random_device()())) noexcept;
```
kevinkreiser: we no longer have a seed method on the class but rather you pass in a sampling function. i personally like the term sample better here but because it has interplay with the "frequency" (nomenclature from statsd proper) i chose to make them match

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -76,14 +107,6 @@ class StatsdClient {
...
     //!@{
 
-    //! Sets a configuration
-    void setConfig(const std::string& host,
-                   const uint16_t port,
-                   const std::string& prefix,
-                   const uint64_t batchsize = 0,
-                   const uint64_t sendInterval = 1000,
-                   const int gaugePrecision = 4) noexcept;
-
```
kevinkreiser, line 86: allowing this makes it possible to break threadsafety gaurantees so its better to just remove it. you can simply construct a new class if you want to configure it differently (should be a rare thing anyway)

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -130,11 +153,8 @@ class StatsdClient {
                 float frequency = 1.0f,
                 const std::vector<std::string>& tags = {}) const noexcept;
 
-    //! Seed the RNG that controls sampling
-    void seed(unsigned int seed = std::random_device()()) noexcept;
-
     //! Flush any queued stats to the daemon
-    void flush() noexcept;
+    void flush() const noexcept;
```
kevinkreiser, line 155: just cleanup

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -152,58 +172,29 @@ class StatsdClient {
 
     //!@}
 
-private:
```
kevinkreiser, line 155: more cleanup

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -152,58 +172,29 @@ class StatsdClient {
...
     //! The prefix to be used for metrics
     std::string m_prefix;
 
     //! The UDP sender to be used for actual sending
     std::unique_ptr<UDPSender> m_sender;
 
-    //! The random number generator for handling sampling
-    mutable std::mt19937 m_randomEngine;
```
kevinkreiser, line 163: not needed as its now encapsulated inside of the random functor below

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -152,58 +172,29 @@ class StatsdClient {
...
-
     //! Fixed floating point precision of gauges
     int m_gaugePrecision;
-};
-
-namespace detail {
```
kevinkreiser, line 169: moved up as previously state

include/cpp-statsd-client/StatsdClient.hpp
```diff
@@ -152,58 +172,29 @@ class StatsdClient {
...
                                   const uint64_t batchsize,
                                   const uint64_t sendInterval,
-                                  const int gaugePrecision) noexcept
+                                  const int gaugePrecision,
+                                  FrequencyFunc  frequencyFunc) noexcept
     : m_prefix(detail::sanitizePrefix(prefix)),
       m_sender(new UDPSender{host, port, batchsize, sendInterval}),
-      m_gaugePrecision(gaugePrecision) {
-    // Initialize the random generator to be used for sampling
-    seed();
```
kevinkreiser, line 195: happens in the declaration as part of the default param

tests/testStatsdClient.cpp
```diff
@@ -53,43 +53,24 @@ void testErrorConditions() {
     throwOnError(client, false, "Should not be able to resolve a ridiculous ip");
 }
 
-void testReconfigure() {
```
kevinkreiser, line 56: functionality is removed so test is removed

tests/testStatsdClient.cpp
```diff
@@ -53,43 +53,24 @@ void testErrorConditions() {
...
-}
-
 void testSendRecv(uint64_t batchSize, uint64_t sendInterval) {
     StatsdServer mock_server;
     std::vector<std::string> messages, expected;
     std::thread server(mock, std::ref(mock_server), std::ref(messages));
 
+    std::mt19937 twister(std::random_device{}());
+    std::uniform_real_distribution<float> dist(0.f, 1.f);
+    auto rand = [&]()->float{return dist(twister);};
```
kevinkreiser: we pass our own custom sampling function so we can fiddle with it to get it to give the results we want

tests/testStatsdClient.cpp
```diff
@@ -159,13 +140,18 @@ void testSendRecv(uint64_t batchSize, uint64_t sendInterval) {
 
     // Make sure we get the exactly correct output
     if (messages != expected) {
-        std::cerr << "Unexpected stats received by server, got:" << std::endl;
-        for (const auto& message : messages) {
-            std::cerr << message << std::endl;
+        for (size_t i = 0; i < expected.size(); ++i) {
```
kevinkreiser, line 143: while i was debugging this trying to get a magical seed that would work for all examples i found changing the formatting of this output makes it easier to debug what is going on

kevinkreiser:
@vthiery i dont know if you have any time to look at this but basically in recent times people are asking about threadsafety again and so this is the last little cleanup needed to make the library fully threadsafe again. more info in #12 and also #53

vthiery:
Thank you for the work @kevinkreiser ! I have to admin that I've only been following this distantly, but I'll review asap.

include/cpp-statsd-client/StatsdClient.hpp
vthiery: Can it be made `const` given that `UDPSender::flush()` is not `const`?

kevinkreiser: its kind of a dirty trick in this case but because the `udpsender` is a `unique_ptr` and its `->` dereference operator is `const` (since it doesnt modify the smart pointer itself) we can then freely call the `non const` member on the dereferenced instance of `udpsender` that the smart ptr returned us

vthiery (approved):

kevinkreiser:
well i saw your ship and so i went to edit the docs to say that the library is thread-safe but upon review i think there is still one thing to do which is to make the sender thread-safe even when the sending interval is `0`. 


see in most use-cases you would make use of the sending interval to spawn the background sending thread and the library is then thread-safe HOWEVER when the sending interval is `0` we expect you to call `flush` manually to send your messages over the socket BUT in this case we acquire a dummy lock (no thread-safety).

therefore if you configure your client with send interval `0` but you share the client in multiple threads who all race to call `flush` you will have thread-safety issues. i'll review the udpsender header further to make sure we remove the rest of the thread-safety pitfalls before asking for another review! apologies!

vthiery (approved):

